PR #19, opened 2025-05-03
Add API to support Discover tab in Freighter

aditya1702:
- Expose API endpoint that reads from the protocols config defined in our kube manifest
- Add unit tests
- Add a custom handler that handles errors thrown by the different API handlers


internal/api/handlers/custom.go
```diff
@@ -0,0 +1,23 @@
...
+// CustomHandler is a wrapper that allows us to process and return errors from different handlers.
+func CustomHandler(f HandlerFunc) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		err := f(w, r)
+		if err != nil {
+			status := http.StatusInternalServerError
+			var apiError *Error
+			if errors.As(err, &apiError) {
+				status = apiError.HttpStatus()
+			}
```
marcelosalloum, line 19: Is the status all you care about in the API error? Are there any extra fields you'd like to render differently?

internal/api/handlers/custom.go
```diff
@@ -0,0 +1,23 @@
...
+			http.Error(w, err.Error(), status)
```
marcelosalloum, line 20: 🤔  Hm, I did a quick research and it seems like the result of `http.Error(...` is not a JSON, but a `text/plain`). Is that want you want here?

internal/api/handlers/error.go
marcelosalloum, line 1: 💡 In the SDP, we added a renderer function to our APIError object. It helps standardizing the response object for API errors: 
https://github.com/stellar/stellar-disbursement-platform-backend/blob/8e84e93fb992f7b00a5ebbd733bf0ca62765ac0e/internal/serve/httperror/httperror.go#L60-L62

internal/api/handlers/health_test.go
```diff
@@ -0,0 +1,29 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestHealthCheckHandler(t *testing.T) {
+	t.Run("should return 200 status code", func(t *testing.T) {
+		t.Parallel()
+		req, _ := http.NewRequest("GET", "/api/v1/health", nil)
```
marcelosalloum, line 15: I know this can look a bit tedious, but asserting all returned fields in unit tests is actually very helpful, as it can often catch some wrong assumptions.

internal/api/handlers/custom.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"errors"
+	"net/http"
+)
+
+type HandlerFunc func(w http.ResponseWriter, r *http.Request) error
+
+// CustomHandler is a wrapper that allows us to process and return errors from different handlers.
+func CustomHandler(f HandlerFunc) http.HandlerFunc {
```
marcelosalloum, line 11: I love this custom handler, great job!

marcelosalloum (commented):
Looking good!

I've left a few comments below, the most important one is related to the error response format apparently being in `text/plain`.

internal/api/handlers/custom.go
aditya1702: Oooh good catch. I did not realise this. I will use the httpjson.Render method that you pointed out to send error as JSON

internal/api/handlers/error.go
aditya1702: This is good. I will use this here

internal/api/handlers/custom.go
aditya1702: I added an extras field too, based on the SDP implementation

marcelosalloum (approved):
LGTM!

PR #20, opened 2025-05-06
Add new error message format

aditya1702:


internal/api/handlers/protocols.go
```diff
@@ -42,28 +59,32 @@ func (h *ProtocolsHandler) GetProtocols(w http.ResponseWriter, r *http.Request)
...
+		return httperror.NewHttpError(ErrFailedToReadProtocolsConfig.ClientMessage, err, http.StatusInternalServerError, nil)
 	}
 
 	var protocols []Protocol
 	err = json.Unmarshal(data, &protocols)
 	if err != nil {
-		errStr := fmt.Sprintf("failed to unmarshal protocols config: %v", err)
-		logger.ErrorWithContext(ctx, errStr)
-		return httperror.NewHttpError(errStr, err, http.StatusInternalServerError, nil)
+		// Helper function to safely get a snippet of data for logging
```
aristidesstaffieri, line 69: Should we lift this helper into a helper file or somewhere that it can be reused by other handlers in the future?

aristidesstaffieri: Yeah just seems like code that will for sure be reused but fine with me to do it later. 👍

PR #26, opened 2025-05-13
Add code coverage to github workflows

aditya1702:
- Add code coverage to github workflow
- Add tests for increasing coverage to greater than 80%


Makefile
```diff
@@ -103,7 +103,7 @@ check: fmt vet lint generate shadow exhaustive deadcode goimports govulncheck ##
 # ==================================================================================== #
 unit-test: ## Run unit tests
 	@echo "==> Running unit tests..."
-	ENABLE_INTEGRATION_TESTS=false go test -v ./...
+	ENABLE_INTEGRATION_TESTS=false go test -v -race -cover -coverprofile=c.out ./...
```
marcelosalloum, line 106: Running the tests with coverage is significantly slower, so I wouldn't make it the default option in your makefile.

marcelosalloum: Perhaps a dedicated makefile function that includes coverage would be better

marcelosalloum (approved):

Makefile
aditya1702: Got it. I will add a separate option fo it

Issue #39, opened 2025-08-20
Support multiple networks from 1 instance


aristidesstaffieri:
### What problem does your feature solve?
Right now, the service is configured to instantiate 1 rpc client per instance which uses the configured RPC url. This means that we can only support 1 network per backend instance.

### What would you like to see?
rpc configuration to happen at the request level so that 1 API instance can answer questions about testnet or pubnet

### What alternatives are there?
Deploy API instance per network

PR #51, opened 2025-10-27
Feature/add ledger key accounts endpoint

piyalbasu:
This is a re-roll of #49 as the scope of this PR has changed significantly. In terms of the feature, this is actually very similar to the old PR. We're now just returning all of the ledger key account's info rather than just the home domain. Besides that, the code is almost all the same, I just changed var names. Please see the old PR for the full discussion.

Closes https://github.com/stellar/freighter-backend/issues/266

_What_

This adds a `/ledger-key/accounts/` endpoint to Freighter Backend V2. This takes an array of public keys, creates Ledger Keys from them, and then passes them on to a single call to RPC to get the account ledger entries of the account(s). From here, we can return account info for each of the requested accounts.

_Why_

Currently in Freighter, when we need to fetch an asset issuer's home domain, we need to look up the account on Horizon. Because of how Horizon accounts are setup, we have to do this one at a time. So, if a user has 10 trustlines, we need to do 10 different Horizon account look ups. This is time consuming. With this new method, we can make one request to Freighter Backend v2 and get all of the relevant home domains in 1 very fast call. 🎉  Also, there's the added benefit of having one less thing that we need Horizon for 🎉 🎉 

Originally, this feature only returned an account's home domain, as that is currently the main use case for Freighter. But after chatting with @aristidesstaffieri , we decided it made more sense for this endpoint to return all of the ledger key's account info. In the future, we may want some other info from the account and it'd be nice to already have this endpoint in place without having to create a new one.

_Future Work_
We still need this endpoint to support Testnet/Futurenet. There is already a [ticket](https://github.com/stellar/freighter-backend-v2/issues/39) for this work. As a fast-follow, I will add RPC network switching and enable it for all of Freighter Backend V2's endpoints

Some examples:

Fetching an account for a valid account on Mainnet:
Screenshot 2025-10-27 at 11 43 54 AM

Fetching an account for multiple valid accounts on Mainnet. The second account does not exist, so it just returns empty values:
Screenshot 2025-10-27 at 11 52 53 AM

Fetching an account for a valid account alongside an invalid account:
Screenshot 2025-10-27 at 12 24 33 PM


internal/api/handlers/ledger_key_accounts.go
```diff
@@ -0,0 +1,177 @@
...
+
+// GetLedgerKeyAccounts handles creating ledger keys from public keys and then fetches the account info from the RPC service
+// It returns a map of public keys to AccountInfo and a map of errors if some of the public keys are invalid
+// This is designed to be flexible so valid public keys will return results while invalid public keys will return errors
+func (h *LedgerKeyAccountHandler) GetLedgerKeyAccounts(w http.ResponseWriter, r *http.Request) error {
+	_, cancel := context.WithTimeout(r.Context(), HealthCheckContextTimeout)
+	defer cancel()
+	ledgerKeyAccountList := make(map[string]types.AccountInfo)
+	var ledgerKeyAccountError LedgerKeyAccountError
+	queryParams := r.URL.Query()
```
aristidesstaffieri: should we provide any validation for missing parameters?

piyalbasu:
@aristidesstaffieri All pr comments addressed here: https://github.com/stellar/freighter-backend-v2/pull/51/commits/b40cca8e33ac59ab706368c84f136dc2304b0e83

PR #52, opened 2025-10-31
add rpc network switching

### LINKED ISSUE #39: Support multiple networks from 1 instance

piyalbasu:
Closes #39 

Support multiple RPC networks. This is needed for multiple endpoints (ledgerkey/accounts and collectibles) to make sure we can fetch data from Mainnet, Testnet, and Futurenet

This adds a network switching configuration and it implements in `collectibles`, `getHealth`, and `ledgerKey/accounts`

### DIFF
--- a/internal/api/handlers/ledger_key_accounts.go
+++ b/internal/api/handlers/ledger_key_accounts.go
@@ -3,11 +3,13 @@ package handlers
 import (
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"maps"
 	"net/http"
-	"slices"
+
+	set "github.com/deckarep/golang-set/v2"
 
 	"github.com/stellar/freighter-backend-v2/internal/api/httperror"
 	response "github.com/stellar/freighter-backend-v2/internal/api/httpresponse"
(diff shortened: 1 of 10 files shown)


internal/api/handlers/collectibles.go
```diff
@@ -243,6 +246,13 @@ func (h *CollectiblesHandler) GetCollectibles(w http.ResponseWriter, r *http.Req
 	ctx, cancel := context.WithTimeout(r.Context(), HealthCheckContextTimeout)
 	defer cancel()
 
+	network := r.URL.Query().Get("network")
+	if network != types.PUBLIC && network != types.TESTNET && network != types.FUTURENET {
```
piyalbasu: In order to support Freighter Mobile which is already using this endpoint in prod, we won't throw an error on a missing network config yet. We'll default to Pubnet to maintain the current functionality

internal/api/handlers/health.go
```diff
@@ -42,10 +43,14 @@ func (h *HealthHandler) CheckHealth(w http.ResponseWriter, r *http.Request) erro
 	w.Header().Set("Cache-Control", "no-cache, no-store, must-revalidate")
 	serviceStatus := make(map[string]string)
 	overallHealthy := true
+	network := r.URL.Query().Get("network")
 
+	if network != types.PUBLIC && network != types.TESTNET && network != types.FUTURENET {
```
piyalbasu: My guess is that this is not being used by a client and that's only being used by us to check health. If that's not accurate, lmk - I'll disable this error

aristidesstaffieri: this gets used in the health check endpoint which is used by the k8s orchestrator so I think we need to do something similar to the collectibles fallback to start, then update our deployment before removing the fallback.

PR #53, opened 2025-11-04
use post for ledger-key/accounts endpoint

piyalbasu:
_What_

Changes the `/ledger-keys/accounts` request method to `POST` instead of `GET`

_Why_

As I was implementing this in the FE, I realized that we may run into instances where an account has a large number of assets for which we will need to fetch home domains. If an account has 100 assets, the request for this endpoint would be over 5600 characters long (each public key being 56 chars). The general rule of thumb is to try to keep a URL limit under 2,000 characters (each browser institutes its own limits).

For scalability, we should err on the side of passing the public keys as JSON body instead of query params

### DIFF
--- a/internal/api/handlers/ledger_key_accounts.go
+++ b/internal/api/handlers/ledger_key_accounts.go
@@ -154,7 +154,7 @@ func (h *LedgerKeyAccountHandler) GetLedgerKeyAccounts(w http.ResponseWriter, r 
 	}
 	ledgerKeyAccountList = ledgerKeyAccountKeys.LedgerKeyAccountMap
 	
-	ledgerKeyAccountsRpcData, e := h.RpcService.GetLedgerEntry(contextWithTimeout, ledgerKeyAccountKeys.LedgerKeys, network)
+	ledgerKeyAccountsRpcData, e := h.RpcService.GetLedgerEntries(contextWithTimeout, ledgerKeyAccountKeys.LedgerKeys, network)
 	
 	if e != nil && ledgerKeyAccountKeysError.ErrorMessage == "" {
 		ledgerKeyAccountError = LedgerKeyAccountError{ErrorMessage: e.Error()}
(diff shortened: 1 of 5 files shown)


aristidesstaffieri:
This will need tweaks to the response headers to match the added method.

internal/api/handlers/ledger_key_accounts.go
```diff
@@ -117,49 +131,46 @@ func processLedgerKeyAccountsEntries(publicKeys []string, data []types.LedgerEnt
...
-			if e != nil && ledgerKeyAccountKeysError.ErrorMessage == "" {
-				ledgerKeyAccountError = LedgerKeyAccountError{ErrorMessage: e.Error()}
-			}
+	ledgerKeyAccountKeys, ledgerKeyAccountKeysError := getLedgerKeyAccountKeys(deduplicatedPublicKeys)
+	if ledgerKeyAccountKeysError.ErrorMessage != "" {
+		ledgerKeyAccountError = ledgerKeyAccountKeysError
+	}
+	ledgerKeyAccountList = ledgerKeyAccountKeys.LedgerKeyAccountMap
+	
+	ledgerKeyAccountsRpcData, e := FetchLedgerEntries(h.RpcService, contextWithTimeout, ledgerKeyAccountKeys.LedgerKeys, network)
```
aditya1702: Shouldnt we have this function as part of the `RpcService`? See how we add the methods to the wallet backend's rpc service: https://github.com/stellar/wallet-backend/blob/main/internal/services/rpc_service.go#L207

piyalbasu: Good point. Actually, after iterations, `FetchLedgerEntries` is now just a simple wrapper around `RpcService.GetLedgerEntry`. I'm going to remove `FetchLedgerEntries` and just directly call `RpcService.GetLedgerEntry`

PR #83, opened 2026-04-17
fix(ledger-key/accounts): max ledger keys for GetLedgerKeyAccounts

aristidesstaffieri:
What
- Cap `public_keys` per request at MaxLedgerKeyAddresses (--max-ledger-key-addresses, default 100) before any decoding or RPC work, mirroring the existing MaxBalanceAddresses pattern.
- Replace the nested loop in `processLedgerKeyAccountsEntries` with an O(N+M) map lookup built from the RPC response.
- Honor `ctx.Err()` inside `getLedgerKeyAccountKeys` so cancellation actually stops the XDR decode loop.
- Add a defensive MaxLedgerEntryKeys (200) guard in `RpcService.GetLedgerEntries` so no future caller can forward an unbounded key slice to the upstream Stellar RPC.

Tests cover over-limit rejection, at-limit success, unlimited mode (max = 0), parity of the new map-based lookup with the old nested loop, and the upstream RPC cap.

Why
`processLedgerKeyAccountsEntries` scanned the RPC entries list once per submitted public key (O(N*M)), and the only input bound was a 1 MB body limit that comfortably fits ~15k keys. A single request could peg a core for ~12 s, and the 5s handler context was never consulted by the loop.

### DIFF
--- a/cmd/serve/serve.go
+++ b/cmd/serve/serve.go
@@ -44,6 +44,7 @@ func (s *ServeCmd) Command() *cobra.Command {
 	cmd.Flags().StringVar(&s.Cfg.AppConfig.MeridianPayStellarHouseAddress, "meridian-pay-stellar-house-address", "", "The Meridian Pay Stellar House collection address")
 	cmd.Flags().Int64Var(&s.Cfg.AppConfig.MaxRequestBodySize, "max-request-body-size", 1<<20, "Maximum request body size in bytes (default: 1MB)")
 	cmd.Flags().IntVar(&s.Cfg.AppConfig.MaxBalanceAddresses, "max-balance-addresses", 100, "Maximum number of addresses allowed in account balances request")
+	cmd.Flags().IntVar(&s.Cfg.AppConfig.MaxLedgerKeyAddresses, "max-ledger-key-addresses", 100, "Maximum number of public keys allowed in a ledger-key/accounts request")
 
 	// RPC Config
 	cmd.Flags().StringVar(&s.Cfg.RpcConfig.PubnetRpcUrl, "pubnet-rpc-url", "", "The Pubnet URL of the Pubnet RPC instance")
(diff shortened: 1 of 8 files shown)


cmd/serve/serve.go
```diff
@@ -44,6 +44,7 @@ func (s *ServeCmd) Command() *cobra.Command {
 	cmd.Flags().StringVar(&s.Cfg.AppConfig.MeridianPayStellarHouseAddress, "meridian-pay-stellar-house-address", "", "The Meridian Pay Stellar House collection address")
 	cmd.Flags().Int64Var(&s.Cfg.AppConfig.MaxRequestBodySize, "max-request-body-size", 1<<20, "Maximum request body size in bytes (default: 1MB)")
 	cmd.Flags().IntVar(&s.Cfg.AppConfig.MaxBalanceAddresses, "max-balance-addresses", 100, "Maximum number of addresses allowed in account balances request")
+	cmd.Flags().IntVar(&s.Cfg.AppConfig.MaxLedgerKeyAddresses, "max-ledger-key-addresses", 100, "Maximum number of public keys allowed in a ledger-key/accounts request")
```
piyalbasu, line 53: Just want to clarify: I believe an operator can set `--max-ledger-key-addresses=10000` at run time, but in  services/rpc.go, we hardcode MaxLedgerEntryKeys to 200. Are these in opposition?

piyalbasu (approved):

cmd/serve/serve.go
aristidesstaffieri: MaxLedgerEntryKeys is meant to be a hard cap so `max-ledger-key-addresses` is meant to be configurable up to that number of ledger keys. I see that this is awkward though and we should probably just enforce the cap in serve.go at startup.

The cap exists to match the max number of ledger keys that RPC will accept for a request.

Moved the validation to startup in 3ab2044e4df8518ad169d356b3a1fe0ab4fceaf9

PR #95, opened 2026-05-05
feat(prices): add /api/v1/token-prices backed by Stellar Expert

aristidesstaffieri:
## Summary

Adds `POST /api/v1/token-prices?network=PUBLIC|TESTNET` to freighter-backend-v2, returning `currentPrice` and `percentagePriceChange24h` per asset for a batch of tokens. Backed by Stellar Expert, fronted by a Redis cache so repeated lookups across users cost zero upstream calls inside a 30s TTL window. Replaces v1's null-only `percentagePriceChange24h` with a real rolling-24h delta derived from hourly candles.

Request body is `{"tokens": [...]}`; the response is keyed by the *original* client input (so a request for `"native"` echoes back `"native"`). Per-token `null` means unpriceable — unknown to Stellar Expert, malformed, or transient upstream error. Futurenet is rejected at the handler.

Scope: pubnet + testnet. Wire-compatible drop-in for the legacy TypeScript `/token-prices` endpoint.

## What's in this PR

- `internal/api/handlers/token_prices.go` — endpoint. Parses + validates input, dedups by canonical asset id, applies the per-request cap *after* dedup (so `[XLM, native, xlm]` collapses to one and isn't rejected on shape). Each id is normalized once and the original→canonical mapping is reused to key the response.
- `internal/services/stellar_expert.go` — thin Stellar Expert client over `/asset/{id}` and `/asset/{id}/candles`. A shared `doJSON` helper maps 404 → `ErrAssetNotFound` and 400 → `ErrAssetMalformed` (distinct from generic upstream errors) so callers map them to nulls without retry. Bearer auth required at startup; `Origin` header is configurable (Cloudflare gates `/candles` on it).
- `internal/services/prices.go` — cache-first orchestration. One Redis MGET per batch; misses fan out via a bounded `errgroup` (`--max-concurrent-price-fetches`, default 25). Concurrent requests for the same cache key coalesce through a process-wide `singleflight.Group`, so a hot key (e.g. XLM) expiring under load triggers a single upstream fetch instead of one per in-flight request. Per-token, asset and candles are fetched concurrently; the candles call is cancelled when the asset returns NotFound/Malformed.
- `internal/services/prices.go` (24h change) — derived solely from the rolling hourly candles window, accepted only when the oldest candle covers ~24h (23–25h guard). When candles are empty, fail to cover the window, or error, `percentagePriceChange24h` is `null` rather than a value mislabeled as 24h (a daily `price7d` series can't pin a trailing 24h). v1 returned `null` here unconditionally.
- `internal/services/prices.go` (cache + degraded modes) — a single Redis TTL (`--price-cache-ttl-seconds`, default 30s): present = hit, absent = miss, with expiry alone governing freshness. On a miss the value is fetched live within `--price-fetch-timeout-seconds` (9s); if that budget is exhausted or the upstream fails, the token comes back `null` best-effort. A cancelled/expired *client* request context propagates as a 503.
- `internal/utils/assetid/` — strict canonicalization of `XLM` / `native` / `CODE:ISSUER`. Rejects garbage at the handler before any cache or upstream work.
- `internal/store/redis.go` — `MGetJSON` / `SetJSON` helpers. Only positive entries are cached.
- `internal/metrics/metrics.go` — three new `freighter_prices_*` Prometheus counters: `cache_outcomes_total{network, outcome}` (hit / miss, per token), `miss_budget_exhausted_total{network}` (timeout flag), `redis_errors_total{op}` (mget / set).
- `cmd/serve/serve.go` + `internal/config/config.go` — config flags (`--stellar-expert-pubnet-url`, `--stellar-expert-testnet-url`, `--stellar-expert-api-key`, `--stellar-expert-origin`, `--price-cache-ttl-seconds`, `--price-fetch-timeout-seconds`, `--max-tokens-per-request`, `--max-concurrent-price-fetches`). Server refuses to start without the API key when token prices are enabled.
- `internal/api/httperror/httperror.go` — drop `originalError` from JSON responses (was emitting `{}` on the wire); the error chain is still preserved in-process via `Unwrap()`.

## Performance vs v1

Measured against `freighter-backend-prd.stellar.org/api/v1/token-prices` with local Redis on 6379, N=8 per cell. v1 serves traffic as-is; v2 Redis is flushed before each cold call.

| Scenario | v1 p50 | v2 cold p50 | v2 warm p50 |
|---|---|---|---|
| 1 token (native) | 126 ms | 180 ms | ~1 ms |
| 4 classic tokens | 129 ms | 187 ms | ~1 ms |
| 5 mixed tokens | 129 ms | 188 ms | ~1 ms |
| Unknown asset | 133 ms | 168 ms | (uncached) |

Warm cache wins ~100×. Cold is +50–60 ms vs v1 — the cost of the second concurrent upstream call (`/candles`) that v1 doesn't make. Cold latency does not grow with token count (per-token concurrency). v2 cold tails are tighter (p90/p50 ~1.0–2.0× vs v1's ~2.7×). Parallelizing asset + candles per token cut cold-cache p50 ~46% (335ms → 180ms) vs the earlier serial implementation. Concurrent cold misses for the same token now collapse to one upstream fetch via singleflight, so a hot key expiring under load no longer multiplies upstream calls.

## Test plan

- [x] Unit + integration tests green locally (`go test ./...`)
- [x] Handler validation matrix, service cache / dedup / concurrency / singleflight-coalescing paths, candles-or-null 24h change, metrics emission, Stellar Expert client (success / 404 / 400 / 5xx / auth header / Origin header / configurable origin / omitted-when-empty), assetid normalization
- [x] `initServices` test confirms server refuses to start without `STELLAR_EXPERT_API_KEY`
- [x] Manual: cold + warm fetches against pubnet from local; `currentPrice` matches v1 prod within <0.4% across native / USDC / yXLM (and v2 adds the rolling 24h change v1 returns `null` for)
- [ ] CI green
- [ ] Staging smoke test before flipping the gate

## Followups

- Short-TTL negative cache for repeatedly-requested unknown assets — each unknown lookup currently re-hits upstream.
- No test exercises a live Redis cache hit (tests use a nil/unreachable store, so every path is a miss); add a miniredis-backed `SetJSON`→`MGetJSON` round-trip + hit-metric test.
- 400 response shape differs from v1's Fastify shape (`{statusCode, code, error, message}` vs v2's `{message, statusCode}` plus optional `extras`). Field names and the set both differ; safe if clients only read `message`/`statusCode`, worth confirming before flipping the gate.


aristidesstaffieri:
@copilot resolve the merge conflicts in this pull request

internal/api/handlers/token_prices.go
```diff
@@ -0,0 +1,106 @@
...
+		return nil, httperror.BadRequest("invalid request body", err)
+	}
+	if len(req.Tokens) == 0 {
+		errStr := "tokens array cannot be empty"
+		return nil, httperror.BadRequest(errStr, errors.New(errStr))
+	}
+	if maxTokens > 0 && len(req.Tokens) > maxTokens {
+		errStr := fmt.Sprintf("too many tokens: maximum is %d, got %d", maxTokens, len(req.Tokens))
+		return nil, httperror.BadRequest(errStr, errors.New(errStr))
+	}
```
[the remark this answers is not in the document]
aristidesstaffieri: deduplication added in 84d88ead1f374ccb07aed5661ea34442797a9001

internal/services/prices.go
```diff
@@ -0,0 +1,500 @@
...
+		defer wg.Done()
+		asset, assetErr = p.stellarExpert.GetAsset(fetchCtx, network, stellarExpertID)
+		if assetErr != nil && (errors.Is(assetErr, ErrAssetNotFound) || errors.Is(assetErr, ErrAssetMalformed)) {
+			cancelFetch()
+		}
+	}()
+	go func() {
+		defer wg.Done()
+		candles, candlesErr = p.stellarExpert.GetAssetCandles(fetchCtx, network, stellarExpertID, from, to, candlesResolutionSec)
+	}()
```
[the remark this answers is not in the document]
aristidesstaffieri: xlm candle skipped in c2f0ccf629cdb9a40990ac81fe22993f0b23072f

internal/api/handlers/token_prices.go
```diff
@@ -0,0 +1,111 @@
...
+		logger.ErrorWithContext(r.Context(), "getting token prices", "error", err)
+		if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled) {
+			return httperror.ServiceUnavailable("token prices temporarily unavailable", err)
+		}
+		return httperror.InternalServerError("Failed to get token prices", err)
+	}
+
+	// Build response keyed by the *original* client input, preserving v1's
+	// echo behavior (so a request for "native" returns "native": ...).
+	out := make(map[string]*types.PriceEntry, len(req.originalInputs))
```
JakeUrban, line 110: Why do we need to support freighter clients using values that aren't considered canonical by the backend?

internal/services/prices.go
```diff
@@ -0,0 +1,507 @@
...
+
+const (
+	pricesServiceName = "prices"
+
+	defaultMaxConcurrent = 25
+	defaultCacheTTL      = 30 * time.Second
+	defaultStaleCacheTTL = 2 * time.Minute
+	defaultMissFetchTTL  = 9 * time.Second
+
+	cacheKeyPrefix = "prices:v1"
```
JakeUrban, line 31: Are we reusing backend-v1's redis cache?

internal/services/prices.go
```diff
@@ -0,0 +1,507 @@
...
+// fetchAndCache hits Stellar Expert for one canonical asset id, writes a
+// successful result to Redis, and returns the response entry. The boolean
+// reports whether the token was authoritatively resolved for this request:
+// not-found/malformed assets resolve to nil; transient failures and budget
+// exhaustion leave the token eligible for stale fallback. The asset and
+// candles calls run concurrently; on a terminal asset error the candles
+// call is cancelled so unknown assets don't double upstream load. Native
+// XLM has no candle data upstream, so the candles call is skipped and the
+// price7d fallback computes the 24h change.
+func (p *pricesService) fetchAndCache(ctx context.Context, network, cacheNet, canonical string) (_ *types.PriceEntry, resolved bool) {
```
JakeUrban: A few questions:

1. Can we get the current price from the candles request?
2. If so, should we make the candles request first, and if the current price or 24hr change isn't available, only then make the asset request? This would limit our requests to 1-per-token in the happy path.
3. Why doesn't XLM have candles?

internal/services/prices.go
aristidesstaffieri: we're not, v1 here is a reference to version of this project's API(api/v1/...)

internal/api/handlers/token_prices.go
aristidesstaffieri: only for parity with the v1 API, we can also drop this and change the client at the time of the migration to send the canonical form for the native token as well. This is the only case in practice that is actually not the canonical form, the asset ID normalization provides a few other conveniences like folding different casings into the same form and trimming white space. If we prefer the API be stricter then we can just reject any non canonical form at the API and update our client.

internal/services/prices.go
aristidesstaffieri: 1. the candles request gives us the close of the previous complete hourly bucket, so its effectively up to 1hr stale for a spot price.
2. I don't think this would actually improve the response times, right now the profile is `max(asset, candles)`, doing it sequentially would make it `candles + maybe asset` which would be slower since we can't get the current price just from that. Additionally to the current price, we need to call `/asset/{id}` to know if a token is price-able, since empty candles could just mean low recent volume.
3. I'm not sure, I'll have to ask the maintainers about this and get back to you.

aristidesstaffieri: there are actually XLM candles, I'm not sure if I made a mistake when querying it the first time or if there was a momentary API blip but I have updated to remove the candle skip for that case in 671dc131035c877ed9e5228131d260f944d3b1c2

aristidesstaffieri:
@copilot resolve the merge conflicts in this pull request

aristidesstaffieri:
This is now deployed in dev, I have spot checked prices against external sources(CEX price for XLM, CoinGecko for Aqua) and our accuracy has improved according to those sources.

```
| Token | prod v1 price | dev v2 price | Stellar Expert direct | CoinGecko |
|---|---|---|---|---|
| XLM | 0.16368007698528217 | 0.16036694383221767 | 0.16037842031484434 | — |
| USDC | 1 (pinned) | 0.9999704457661734 | 0.9999704638245943 | — |
| AQUA | 0.00034565995663034683 | 0.00034380487473591317 | 0.00034380964552642955 | 0.00034667 |
| yXLM | 0.16361628376427706 | 0.1599375891550416 | 0.16000494229962905 | — |
```

USDC is hardcoded in v1, we can see the true price in v2 according to Stellar Expert.

internal/api/handlers/token_prices.go
JakeUrban: I don't have a strong preference, up to you.

aristidesstaffieri: ok I think I will leave it as is in order to make the UI cut over as seamless as possible.

configs/.toml-EXAMPLE
```diff
@@ -32,6 +32,17 @@ USE_BLOCKAID_TRANSACTION_WARNING_REPORTING = "not-set"
 COINBASE_API_KEY = "not-set"
 COINBASE_API_SECRET = "not-set"
 
+# Token Prices
+STELLAR_EXPERT_PUBNET_URL = "not-set"
+STELLAR_EXPERT_TESTNET_URL = "not-set"
+# Required. Server refuses to start without it.
```
aditya1702: We have defined default values in the flag config so wont the server start even if these are not specified?

internal/services/prices.go
```diff
@@ -0,0 +1,501 @@
...
+}
+
+func (p *pricesService) recordCacheOutcome(network, outcome string, n int) {
+	if p.pricesMetrics == nil || n <= 0 {
+		return
+	}
+	p.pricesMetrics.CacheOutcomes.WithLabelValues(network, outcome).Add(float64(n))
+}
+
+func (p *pricesService) resolveMisses(ctx context.Context, network, cacheNet string, misses []string, result map[string]*types.PriceEntry, resultMu *sync.Mutex) {
```
aditya1702, line 193: Two things on the miss fan-out:

1. **No cross-request coalescing.** Concurrent requests for the same token each run their own `GetAsset` + `GetAssetCandles`, so when a hot key like XLM expires every in-flight request fans out to upstream (~2N calls where 2 would do), worst for the most-requested tokens. With Orbit about to assess our usage to set a rate limit, that inflates our numbers. Suggest wrapping the per-token fetch in an in-process `singleflight.Group` keyed by the cache key.
2. **Hand-rolled pool.** This reimplements a bounded worker pool (jobs channel + `WaitGroup` + per-worker `ctx.Done()` select + mutex). `wallet_backend.go:192` already does the identical bounded fan-out with `errgroup.WithContext` + `g.SetLimit(...)` (and `x/sync` is already a dep) — reusing that idiom would cut this ~40 lines to ~12 with cleaner cancellation.

internal/services/prices.go
```diff
@@ -0,0 +1,501 @@
...
+	}
+	p.cachePositive(ctx, cacheNet, canonical, entry)
+	return entry, true
+}
+
+// changeFromCandlesOrFallback picks the best 24h-change source: the candles
+// window when it covers ~24h with a non-zero open, otherwise the daily
+// price7d fallback. Candles errors that aren't context-related fall back to
+// price7d after a single warning log.
+func changeFromCandlesOrFallback(asset *types.StellarExpertAsset, candles []types.StellarExpertCandle, candlesErr error, stellarExpertID string, to time.Time) *string {
```
aditya1702: Suggest dropping the `price7d` fallback and making the 24h change candles-or-`null`:

- **Correctness:** the requirement is a *trailing*-24h change. `price7d` is a daily UTC-midnight series, so it can't pin "now − 24h" — and `compute24hChange` compares the live price to `dailyCandles[len-2]` (~48h old), so the fallback is labelled 24h but isn't. `change24hFromCandles` already returns `nil` cleanly when candles can't cover ~24h, and `null` is a supported outcome (v1 returned `null` here unconditionally).
- **Cascade** — removing it deletes: `compute24hChange`, the `maxPrice7dFallbackAge` const + comment, the `StellarExpertAsset.Price7d` field + decode (now unused), and collapses this function to "candles error → `nil` (warn-log non-context errors); else → `change24hFromCandles`" (thin enough to inline into `fetchAndCache`). The duplicated rounding/neg-zero/`FormatFloat` tail also disappears on its own — no shared helper needed.

Keep `change24hFromCandles` and its 23–25h window guard as-is; that's the part doing the correct work.

internal/services/prices.go
```diff
@@ -0,0 +1,501 @@
...
+
+type cacheFreshness int
+
+const (
+	cacheMissing cacheFreshness = iota
+	cacheFresh
+	cacheStale
+)
+
+func (c *cachedPriceEntry) freshness(now time.Time, freshTTL, staleTTL time.Duration) cacheFreshness {
```
aditya1702: The two-tier stale-if-error only helps a narrow case — a token re-requested within the 30–120s band whose refetch *also* fails — and it only postpones `null` by ≤90s, so it doesn't survive a real outage; the fresh 30s cache already absorbs brief blips. Serving a silently-stale price (no staleness signal) is also arguably worse than `null` for a widget that already treats `null` as normal.

Suggest a single configurable TTL (keep `--price-cache-ttl-seconds`) and `null` on refetch failure. Removes `StaleCacheTTL` + the `--price-stale-cache-ttl-seconds` flag + its `stale >= fresh` check in serve.go, the `cacheFreshness` enum + `freshness()`, the `FetchedAt` field + per-read parse, the `staleFallback` map, the stale branch of `completeMissingResults`, and the `stale_fallback_served` metric. If we want real outage resilience later, a *long* last-good window (decoupled from the fresh TTL) is the deliberate way to add it.

internal/types/interfaces.go
```diff
@@ -41,5 +42,41 @@ type RPCService interface {
...
+// we care about for pricing.
+type StellarExpertAsset struct {
+	Price   float64      `json:"price"`
+	Price7d [][2]float64 `json:"price7d"`
+}
+
+// StellarExpertCandle is one row of /asset/{id}/candles.
+// Wire shape per Stellar Expert API docs: [ts, open, low, high, close,
+// quote_volume, base_volume, trades]. Native XLM returns no candles.
+type StellarExpertCandle [8]float64
```
aditya1702, line 59: Two small cleanups on this type:

- **`Close()` (line 63) is unused in production** — the 24h calc only reads `TS()`/`Open()`, and the array decodes fine without it. Suggest removing it (and its one test reference); re-add when something consumes the close price.
- **The "Native XLM returns no candles" comment (line 58) looks stale** — `fetchAndCache` fetches XLM candles unconditionally, and the skip was reverted in `671dc13` ("remove unfounded XLM candles skip"). The PR description still says "Native XLM skips the candles call" too. Worth reconciling: if XLM does return candles (the revert implies it does), fix the comment + PR description.

internal/api/handlers/token_prices.go
```diff
@@ -0,0 +1,111 @@
...
+	for _, original := range req.originalInputs {
```
aditya1702, line 111: `assetid.Normalize` runs twice per token — once in `validateTokenPricesRequest` and again here in the response loop (error swallowed, only safe because validation already passed). Suggest having validation return an `original → canonical` mapping and indexing that here, so we normalize once and skip ~1000 redundant ed25519 decodes on a max-size request.

internal/services/prices.go
```diff
@@ -0,0 +1,501 @@
...
+	defaultStaleCacheTTL = 2 * time.Minute
+	defaultMissFetchTTL  = 9 * time.Second
+
+	cacheKeyPrefix = "prices:v1"
+
+	// MaxTokensPerPriceRequest is the hard ceiling for --max-tokens-per-request.
+	// One request fans out one upstream Stellar Expert call per cache miss, so
+	// the cap bounds the per-request amplification factor regardless of how
+	// the operator sets the runtime flag.
+	MaxTokensPerPriceRequest = 1000
```
aditya1702: We have both a `--max-tokens-per-request` flag (default 1000) and this `MaxTokensPerPriceRequest = 1000` compile-time ceiling validating it (serve.go:36). Is the ceiling a deliberate hard safety bound on per-request amplification regardless of the flag? If so a one-line note would clarify intent; if it's just guarding the flag default, one source of truth (flag *or* constant) is simpler than maintaining both plus the `PreRunE` check and the two ceiling tests.

aditya1702 (commented):
Did a thorough design + simplification pass. Overall this is solid, standard work — a read-through Redis cache over the Stellar Expert API with bounded fan-out is the right shape, and using `/candles` for a true trailing-24h is the correct call. A few things to work through before merge, plus simplifications that cut a good chunk of code without losing behavior.

Themes:
> - **Concurrency** — no cross-request coalescing on misses; the fan-out can reuse our existing `errgroup` idiom.
> - **24h change** — drop the `price7d` fallback (can't represent trailing-24h, and the math is off); deletes a lot of code.
> - **Caching** — collapse the two-tier stale cache to a single TTL.
> - **Dead/duplicated code** — unused `Close()` + a stale XLM comment, double `Normalize`, and a redundant max-tokens ceiling.

internal/services/prices.go
aristidesstaffieri: as we discussed, I collapsed the ttls down to 1, cache freshness is now just a product of cache hits or misses.
96819f331738dba6ccc2ee45594755e9afd24a3f

configs/.toml-EXAMPLE
aristidesstaffieri: yes true, removed this comment in 50b22ff5bd9b6286dbd359c042edde2c984493cc

internal/services/prices.go
aristidesstaffieri: added request coalescing and replaced the pool with a channel + wait group worker pool in 50b22ff5bd9b6286dbd359c042edde2c984493cc

internal/api/handlers/token_prices.go
aristidesstaffieri: I refactored this to normalize each token ID only once in 50b22ff5bd9b6286dbd359c042edde2c984493cc

internal/types/interfaces.go
aristidesstaffieri: fixed both of these issues in 50b22ff5bd9b6286dbd359c042edde2c984493cc

internal/services/prices.go
aristidesstaffieri: yeah we can just trust that the operator will set the flag sanely, I dropped the hard limit in 5f655684183e7e6e80758022f6f13dc497cb798b

internal/services/prices.go
aristidesstaffieri: yeah fair enough, we can accept the nulls for tokens with a sparse price history and that will match our current implementation and be simpler. I removed the 7 day fallback in 7f6f19235f2a42c46a8fd2fa7fb90764d48e965c

aditya1702 (approved):

PR #96, opened 2026-05-06
feat(ci): add prerelease publish + promote-to-release workflows

aditya1702:
Closes #69 

## Summary
- Adds `publish-prerelease.yml` — workflow_dispatch, builds & pushes a prerelease image from `main` HEAD to the `stg/freighter-backend-v2` namespace of the internal ECR registry, creates a GitHub prerelease.
- Adds `promote-release.yml` — workflow_dispatch, re-tags the staging image into the `prd/freighter-backend-v2` namespace (digest-preserving via `--prefer-index=false`, no rebuild), creates a non-prerelease GitHub release.
- Documents the release flow in `README.md`.

## Design
- Tag format: `vX.Y.Z` for releases, `vX.Y.Z-rc.N` for prereleases. Strict.
- Docker tag mirrors git tag verbatim (no `rc-` prefix).
- No `:latest` tag — kube manifests pin to explicit versions.
- Releases are always cut from `main` HEAD (no `target_ref` input — closes a privilege-escalation hole where feature-branch Makefile/Dockerfile would otherwise run with ECR push credentials).
- Promotion uses `docker buildx imagetools create --prefer-index=false` so the source manifest is carbon-copied (same digest) rather than wrapped in a new OCI index.
- GitHub release is only created after the image step succeeds — a failed push never produces an orphan release.

## Prerequisites (out of this PR)
- ECR repos `stg/freighter-backend-v2` and `prd/freighter-backend-v2` must exist in the internal SDF ECR registry.

(The `stellar/actions/sdf-ecr-login@main` action hardcodes its OIDC and ECR-push role ARNs internally, so no repo-level role secrets are needed.)

## Test plan
- [x] `actionlint` clean on both new files
- [ ] After merge: dispatch `publish-prerelease` with `version=v0.0.0-rc.0` → verify image in stg ECR + GitHub prerelease
- [ ] After merge: dispatch `promote-release` with `prerelease_tag=v0.0.0-rc.0`, `release_version=v0.0.0` → verify identical digest in prd ECR + non-prerelease GitHub release
- [ ] Cleanup: delete `v0.0.0-rc.0` and `v0.0.0` releases/tags + ECR images after validation

## Why workflow_dispatch and not release-event-triggered?
- Both workflows create their own GitHub releases after image steps succeed. The user's single action is "trigger the workflow" — no separate `gh release create` step that could be forgotten or misordered.
- This is symmetric (both publish and promote work the same way) and keeps side effects atomic per workflow run.

## Codex review fixes
- `--prefer-index=false` to preserve manifest digest (single-platform source).
- Remove `target_ref` input — privilege-escalation mitigation.
- Remove unused `with:` block on ECR login (the action ignored those inputs).

### DIFF
--- a/.github/workflows/promote-release.yml
+++ b/.github/workflows/promote-release.yml
@@ -117,3 +117,14 @@ jobs:
             exit 1
           fi
           echo "Promotion successful — bytes preserved."
+
+      - name: Create GitHub release
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RELEASE_VERSION: ${{ inputs.release_version }}
+          COMMIT_SHA: ${{ steps.resolve_sha.outputs.sha }}
+        run: |
+          gh release create "$RELEASE_VERSION" \
+            --target "$COMMIT_SHA" \
+            --title "$RELEASE_VERSION" \
+            --generate-notes
--- a/.github/workflows/promote-release.yml
+++ b/.github/workflows/promote-release.yml
@@ -93,9 +93,6 @@ jobs:
       - name: ECR login via OIDC
         id: ecr_login
         uses: stellar/actions/sdf-ecr-login@main
-        with:
-          aws-oidc-role: ${{ secrets.AWS_GITHUB_OIDC_ROLE }}
-          aws-ecr-login-role: ${{ secrets.AWS_ECR_LOGIN_ROLE }}
 
       - name: Re-tag image from staging to production ECR
         env:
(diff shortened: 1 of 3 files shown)


aristidesstaffieri:
Follow-up on `publish-prerelease.yml`: this workflow does not reserve `inputs.version` until the final `gh release create`, but it pushes `stg/freighter-backend-v2:$VERSION` earlier. If the same version is dispatched twice, or a failed run is retried after `main` changes, both runs can pass the tag-existence check and race to overwrite the staging image tag. The GitHub release/tag will end up pointing at the commit from one run, but the staging image tag may already have been overwritten by the other digest.

I think we should add a per-version `concurrency` group here and, as defense in depth, fail before push if the target ECR tag already exists.

PR #98, opened 2026-05-07
feat(account-balances): fan out per-address fetches via `GetAllAccountBalances` in wallet-backend SDK

### LINKED ISSUE #97: Use goroutines to fetch balances for multiple accounts from wallet-backend
Wallet-backend had a multi-account balance endpoint `BalancesByAccountAddresses` that fetched balances for multiple addresses at once. However we removed this to keep the GraphQL API in sync with best practices and now the only way to get a balance is to query the balance fields through `accountByAddress` query.

Since the freighter-backend-v2's account balances endpoint is a passthrough to wallet backend, we have to implement the multi-account logic here using goroutines that calls the WB's single account balance endpoint.

aditya1702:
Closes #97 

## Summary

- Bumps wallet-backend SDK to the post-#612 commit that exposes `client.GetAllAccountBalances` (auto-paginated single-account, flat slice) and the `wbclient.ErrAccountNotFound` typed sentinel. The multi-account `BalancesByAccountAddresses` GraphQL endpoint was removed upstream and is rebuilt freighter-side as a per-request fan-out.
- Rewrites `walletBackendService.GetBalancesByAccountAddresses` to fan out one `GetAllAccountBalances` call per unique requested address with bounded per-request concurrency via `errgroup.SetLimit`. The SDK owns per-account pagination internally.
- Splits failures into address-scoped (per-account `Error` string, request returns 200) vs systemic (top-level 5xx), so monitoring still sees outages instead of a 200 of error strings.
- Preserves the historical wallet-backend resolver dedupe behavior (collapse duplicates, keep first-seen order).
- Renames the route from `POST /api/v1/account-balances` to `POST /api/v1/accounts/balances` to follow REST resource-oriented conventions (safe hard-rename — endpoint is undeployed).

## Why

Wallet-backend removed both `BalancesByAccountAddresses` (multi-account) and `BalancesByAccountAddress` (single-account) endpoints in favor of a paginated `accountByAddress { balances { ... } }` query. The freighter-backend handler is a passthrough, so multi-account aggregation now has to happen here.

## Behavior changes

- **Wire format**: the handler still passes SDK structs through to JSON. The SDK upgrade itself changes the inner balance shape — new `NativeBalance` adds `minimumBalance`, `liabilities`, and `lastModifiedLedger`; `SEP41Balance` is gone. Consumers must accommodate the new shape.
- **Per-account `Error`**: response items now carry an optional `error` string for address-scoped failures (`graphql_error`, `account_not_found`). Re-defined freighter-side because the SDK's `AccountBalances` struct was deleted with the multi-account endpoint.
- **Address-scoped policy**: `graphql_error` and `account_not_found` (the new typed sentinel `wbclient.ErrAccountNotFound`, signaled by a 200 with `accountByAddress: null`) are address-scoped. **Every `http_error` is now systemic** — including 4xx. wbclient POSTs every call to `/graphql/query` and translates any HTTP `>= 400` into an `http_error` before parsing the GraphQL body, so a 4xx from upstream means a wrong base URL, missing route, auth/signing failure, rate limit, or schema-violating query — never "this account doesn't exist." Failing the whole batch on 4xx lets monitoring catch real outages instead of returning a 200 with N per-account error strings.
- **Empty balances**: an account with zero balances marshals as `"balances": []` (never `null`). The SDK returns a non-nil empty slice; the service keeps a defensive guard so a future SDK regression to nil wouldn't leak `"balances": null` into the wire format.
- **Pagination**: the SDK's `GetAllAccountBalances` drives the per-account page loop with `first=100` and includes pagination-loop defenses (errors on missing `EndCursor` when `HasNextPage=true`, errors on a stuck cursor). Freighter-backend doesn't expose per-account pagination on its own API today; if/when it does, swap to `GetAccountBalances` (one page) at that boundary.
- **Route rename**: `POST /api/v1/account-balances` → `POST /api/v1/accounts/balances`. Old path returns 404.

## Configuration

New flag: `--wallet-backend-balance-concurrency` (default `10`). Caps the per-request goroutine count for the fan-out. Startup rejects non-positive values to avoid the `errgroup.SetLimit(0)` (blocks forever) and `SetLimit(-1)` (unbounded) surprises.

## Test plan

- [x] `make check` (excluding `exhaustive`, which fails pre-existing on `main` due to a Go 1.26 / `golang.org/x/tools v0.15.0` incompatibility — unrelated)
- [x] `make unit-test` — `TestGetBalancesByAccountAddresses_FanOut` covers: success ordering, multi-page aggregation via the SDK loop, per-account graphql error, per-account `account_not_found`, systemic 4xx (404 fails the whole request), systemic 5xx, transport errors, zero-balances normalization, bounded concurrency with a barrier, request-level timeout, and dedupe. `TestIsAddressScopedError` covers explicit per-code rows for 400/401/403/404/422/429/503 (all systemic) plus `graphql_error` and `account_not_found` (address-scoped).
- [ ] Manual smoke test against a reachable wallet-backend (`make docker-build-up`, then `curl -X POST 'http://localhost:3002/api/v1/accounts/balances?network=PUBLIC' -d '{"addresses":["G..."]}'`)
- [ ] Verify metrics on `localhost:9090/metrics` — `freighter_service_calls_total{service="wallet-backend",method="GetBalancesByAccountAddresses"}` increments and `freighter_service_call_duration_seconds` reports reasonable latency.

## Out of scope

- No integration test container for wallet-backend (existing setup doesn't have one; explicitly authorized to skip).
- No per-address Prometheus metrics — keeping aggregate-only.
- No freighter-owned DTOs for the inner balance shape — passthrough preserved by design.
- No caching layer for balance responses.
- No typed-error hierarchy for the rest of the upstream classifier (`classifyWBError` still substring-matches for `"GraphQL error:"` and `"unexpected statusCode="`). PR #612 demonstrated the pattern works for `ErrAccountNotFound`; extending to auth/rate-limit/transport is a wallet-backend SDK follow-up.


piyalbasu:
### Code review

Found 1 issue:

1. The godoc for `GetBalancesByAccountAddresses` names the wrong SDK method — it says "`wbclient.GetAccountBalances`" but the implementation calls `client.GetAllAccountBalances(gctx, addr)` (the paginating variant). Worth fixing since the test file explicitly warns against swapping back to the non-paginating call.

https://github.com/stellar/freighter-backend-v2/blob/5a9c4f424ec319ad8d6b2808aa17208087ce4ad5/internal/services/wallet_backend.go#L147-L198

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

cmd/serve/serve.go
```diff
@@ -53,6 +59,7 @@ func (s *ServeCmd) Command() *cobra.Command {
 	cmd.Flags().Int64Var(&s.Cfg.AppConfig.MaxRequestBodySize, "max-request-body-size", 1<<20, "Maximum request body size in bytes (default: 1MB)")
 	cmd.Flags().IntVar(&s.Cfg.AppConfig.MaxBalanceAddresses, "max-balance-addresses", 100, "Maximum number of addresses allowed in account balances request")
 	cmd.Flags().IntVar(&s.Cfg.AppConfig.MaxLedgerKeyAddresses, "max-ledger-key-addresses", 100, "Maximum number of public keys allowed in a ledger-key/accounts request")
+	cmd.Flags().IntVar(&s.Cfg.AppConfig.WalletBackendBalanceConcurrency, "wallet-backend-balance-concurrency", 10, "Per-request maximum number of concurrent wallet-backend balance fetches (the /account-balances handler fans out to one accountByAddress call per address)")
```
piyalbasu: Should we set an upper limit here? In practice, probably not an issue in practice, but might be something to consider

from claude:
```
All goroutines share one http.Client (wallet_backend.go#L50-L52). The Transport's MaxConnsPerHost / MaxIdleConnsPerHost will throttle real concurrency — excess goroutines block waiting
  for a connection slot. Worth checking what's set; if it's the Go default (MaxIdleConnsPerHost=2, MaxConnsPerHost=0/unlimited), you can open arbitrarily many sockets to wallet-backend.
```

aditya1702: Good catch on raising it — fwiw the AI quote in the comment assumed Go defaults. The shared `http.Client` we hand to `wbclient` (`internal/services/wallet_backend.go:50-63`) sets `MaxConnsPerHost: 50`, `MaxIdleConnsPerHost: 10`, and a 30s client timeout, so excess goroutines block in `Transport.getConn` waiting for a slot rather than opening more sockets — the real upper bound on concurrency is the transport, not this flag.

The remaining footgun is goroutine memory if someone sets the flag to something absurd like 100k, but that's bounded (~2-8KB per blocked goroutine) and recoverable. Not worth a hard ceiling IMO — leaving as-is.

aristidesstaffieri:
### Code review

Found 2 issues:

1. Three comment strings still reference the old endpoint path `/api/v1/account-balances` after the rename to `/api/v1/accounts/balances` in commit 06e781b.

- `internal/types/wallet_backend.go` ABOUTME line 1: \`// ABOUTME: Response types for the wallet-backend-fronted /api/v1/account-balances endpoint.\`
- `internal/config/config.go` field comment line 29: \`// fetches per single /api/v1/account-balances request.\`
- `cmd/serve/serve.go` flag help string line 62: \`"...the /account-balances handler fans out..."\`

https://github.com/stellar/freighter-backend-v2/blob/06e781b766f49a9cd344372e9a73d3b1237364b6/internal/types/wallet_backend.go#L1

https://github.com/stellar/freighter-backend-v2/blob/06e781b766f49a9cd344372e9a73d3b1237364b6/internal/config/config.go#L28-L30

https://github.com/stellar/freighter-backend-v2/blob/06e781b766f49a9cd344372e9a73d3b1237364b6/cmd/serve/serve.go#L62

2. The godoc for `GetBalancesByAccountAddresses` (line 148) names the wrong SDK method — it says `wbclient.GetAccountBalances` but the implementation calls `client.GetAllAccountBalances`. The test file explicitly guards against swapping back to the non-paginating form (`multi_page_account_aggregates_all_pages_via_sdk_loop`), making the mismatch in the godoc misleading for the next reader.

https://github.com/stellar/freighter-backend-v2/blob/06e781b766f49a9cd344372e9a73d3b1237364b6/internal/services/wallet_backend.go#L148-L151

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

aditya1702:
Fixed issue 1 in 302c10c — three stale `/api/v1/account-balances` references in comments/help text updated.

Skipped issue 2: the godoc at `internal/services/wallet_backend.go:148` already reads `wbclient.GetAllAccountBalances`, not `wbclient.GetAccountBalances` as quoted in the review — no change needed there.

aristidesstaffieri (approved):

PR #91, opened 2026-04-29
fix(collectibles): sanitize user-facing error messages 

aristidesstaffieri:
Closes #84 

What
Sanitize and use end user facing error messages for collectibles route.

Why
Internal RPC/transport errors (e.g. "dial tcp <IP>: connect: connection refused") were being returned in the error_message field of collectibles responses, leaking infrastructure details to clients. Replace these with generic user-facing messages while logging the underlying error server-side for debugging, and add a regression test that verifies no internal details leak through.

PR #100, opened 2026-05-13
feat: account-history — single transactions endpoint with embedded ops + state changes

### LINKED ISSUE #101: Add history endpoints to query an account's transactions, operations and state changes
The new history redesign will use wallet-backend's graphql endpoints to get an account's transactions, operations and state changes. We want to add the relevant REST endpoints on backend v2 that will passthrough to wallet backend using the Go client.

aditya1702:
Closes #101 

## Summary

Collapses the three account-history endpoints into a single call:

- **`GET /api/v1/accounts/{address}/transactions`** — paginated transactions, each embedding the calling account's `operations` and `state_changes`.

Backed by the new wallet-backend SDK method `GetAccountTransactionsWithOpsAndStateChanges` (one nested GraphQL call — ~3 batched DB queries per page, not N+1). The response is fully **snake_case** with **narrow-polymorphic** state changes (discriminated by `type`) and 64-bit `id`/`fee_charged` encoded as JSON **strings** for JS-client precision.

## Changes
- `internal/types/account_history.go` — snake_case response types: `AccountTransaction` (embeds base `Transaction`), `Operation`, and the `StateChange` interface + 9 variants.
- `internal/services/account_history_mapping.go` — pure wbclient→snake_case mappers (incl. the 9-way state-change switch).
- `internal/services/wallet_backend.go` — `GetAccountTransactions` rewritten to the single nested call; the two flat service methods + empty-page helpers removed.
- `internal/types/interfaces.go`, `internal/utils/mocks.go`, `internal/api/handlers/account_history.go`, `internal/api/serve.go` — trimmed to the one method / one route.
- `cmd/serve/serve.go` — `account-history-max-limit` capped at 100 (the upstream page-size ceiling), failing fast at startup instead of 502-ing per request.

### DIFF
--- a/internal/api/handlers/utils.go
+++ b/internal/api/handlers/utils.go
@@ -17,6 +17,15 @@ func isValidNetwork(network string) bool {
 	return network == types.PUBLIC || network == types.TESTNET || network == types.FUTURENET
 }
 
+// isValidWalletBackendNetwork reports whether network is one of the values
+// freighter-backend has a wallet-backend client configured for. WalletBackendConfig
+// only carries pubnet and testnet URL/signing-key fields, so FUTURENET (which is
+// otherwise a valid Stellar network in this codebase) is intentionally rejected
+// to prevent silent fall-through to the pubnet client.
+func isValidWalletBackendNetwork(network string) bool {
+	return network == types.PUBLIC || network == types.TESTNET
+}
+
 type collection struct {
 	Name   string `json:"name"`
 	Symbol string `json:"symbol"`
--- a/internal/api/handlers/utils.go
+++ b/internal/api/handlers/utils.go
@@ -2,13 +2,20 @@ package handlers
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"net"
+	"net/url"
 	"strconv"
 
 	"github.com/alitto/pond/v2"
 	"github.com/stellar/go-stellar-sdk/txnbuild"
 	"github.com/stellar/go-stellar-sdk/xdr"
+	"github.com/stellar/wallet-backend/pkg/wbclient"
 
+	"github.com/stellar/freighter-backend-v2/internal/api/httperror"
+	"github.com/stellar/freighter-backend-v2/internal/logger"
+	"github.com/stellar/freighter-backend-v2/internal/metrics"
 	"github.com/stellar/freighter-backend-v2/internal/types"
 	"github.com/stellar/freighter-backend-v2/internal/utils"
 )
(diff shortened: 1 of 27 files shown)


aristidesstaffieri:
### Code review — issue 1 of 3

**`context.Canceled` is mapped to HTTP 504 with a "timed out" log message.**

`translateServiceError` lumps `context.DeadlineExceeded` and `context.Canceled` together, returning `GatewayTimeout` (504) and logging `"wallet-backend call timed out"`. But `Canceled` fires on client disconnect / parent-request abort — it is not a gateway timeout. Since this function is shared by balances and all three new account-history endpoints, the misclassification propagates broadly into metrics and logs.

Prior precedent in this repo points the other way:
- PR #91 commit `d5e2849` explicitly removed a "Request timed out" label from the collectibles handler with the reason: *"ctx.Err() can be context.Canceled as well, making the message inaccurate."*
- PR #98 commit `4137f6f` deliberately mapped both `DeadlineExceeded` and `Canceled` to **503** (not 504) in the prices handler.

Suggest splitting the cases: return 504 only for `DeadlineExceeded`, and either 503 (matching `prices.go`) or no response (client already gone) for `Canceled`. At minimum the log message should not say "timed out" on the `Canceled` branch.

https://github.com/stellar/freighter-backend-v2/blob/d6f343b1e13d756510181e24a33417fde8034e47/internal/api/handlers/utils.go#L44-L48

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

aristidesstaffieri:
### Code review — issue 2 of 3

**Startup validation misses the `MaxInt32` ceiling — operator misconfig panics inside `Start()`.**

The new `PersistentPreRunE` check validates `d <= 0 || m <= 0 || d > m` but does not check `m > math.MaxInt32`. Meanwhile `NewAccountHistoryHandler` does enforce the `MaxInt32` cap (commit `2065e21`, CodeQL fix), and `initHandlers()` panics on the constructor error. Net effect: an operator setting `--account-history-max-limit=3000000000` (a valid positive `int` above `MaxInt32`) passes PreRunE and then crashes mid-`Start()` instead of failing fast at flag parse time.

Suggest mirroring the constructor's cap in PreRunE so startup validation is the single source of truth. Prior PRs adding similar startup caps (e.g. `aec3a20`, `7b1c047` for `--max-ledger-key-addresses`) followed this pattern, including matching tests in `cmd/serve/serve_test.go`.

PreRunE check:
https://github.com/stellar/freighter-backend-v2/blob/d6f343b1e13d756510181e24a33417fde8034e47/cmd/serve/serve.go#L34-L39

Panicking init path:
https://github.com/stellar/freighter-backend-v2/blob/d6f343b1e13d756510181e24a33417fde8034e47/internal/api/serve.go#L113-L120

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

aristidesstaffieri:
### Code review — issue 3 of 3

**`initHandlers()` panics on constructor error, contradicting the stated "matching `NewWalletBackendService`'s pattern".**

The godoc on `NewAccountHistoryHandler` says it returns an error "so `cmd/serve` fails fast at startup, matching `NewWalletBackendService`'s pattern." But `NewWalletBackendService`'s error is propagated cleanly through `initServices() error` up to `Start()`, while `initHandlers()` returns only `*http.ServeMux` — leaving no choice but `panic(...)` on the new constructor error.

In practice the panic path is dead because `PersistentPreRunE` already validates the inputs (and issue 2 above would close the remaining gap). But the contract stated in the godoc is broken. Suggest changing `initHandlers()` to `(*http.ServeMux, error)` and returning the error up through `Start()`.

https://github.com/stellar/freighter-backend-v2/blob/d6f343b1e13d756510181e24a33417fde8034e47/internal/api/serve.go#L113-L120

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

internal/api/handlers/utils.go
```diff
@@ -2,19 +2,64 @@ package handlers
...
+}
+
+// translateServiceError maps a service-layer error to a typed HttpError per
+// the spec's REST-honest mapping. Logs all non-404 errors with context;
+// account-not-found is a normal client outcome and is not logged.
+//
+// Status mapping:
+//   - wbclient.ErrAccountNotFound       -> 404
+//   - context.DeadlineExceeded/Canceled -> 504
+//   - *metrics.UpstreamError (any Kind) -> 502 (graphql_error, http_error)
```
JakeUrban, line 40: Why is this error defined on the metrics package?

JakeUrban (approved):
LGTM, but @aristidesstaffieri should also review.

internal/api/handlers/utils.go
aditya1702: @JakeUrban I think this was added previously by Piyal/Aristides - dont remember now exactly. But yeah it should not be in metrics package

aditya1702:
@aristidesstaffieri thanks for the thorough review — all three issues addressed in separate commits:

- **Issue 1 (`context.Canceled` → 503)** — fixed in [`bacda4b`](https://github.com/stellar/freighter-backend-v2/commit/bacda4b). Split `Canceled` from `DeadlineExceeded` in `translateServiceError`; `Canceled` now returns 503 `ServiceUnavailable` with a distinct "canceled by client" log message, matching the `prices.go` precedent in `4137f6f`.
- **Issue 2 (PreRunE missing `MaxInt32` ceiling)** — fixed in [`414438e`](https://github.com/stellar/freighter-backend-v2/commit/414438e). Added `m > math.MaxInt32` to the `PersistentPreRunE` guard with a matching test, so an out-of-range `--account-history-max-limit` fails fast at flag parse instead of panicking inside `Start()`.
- **Issue 3 (`initHandlers()` panic vs godoc contract)** — fixed in [`30b4da6`](https://github.com/stellar/freighter-backend-v2/commit/30b4da6). `initHandlers()` now returns `(*http.ServeMux, error)` and the error propagates through `Start()` the same way `initServices()` does — no more panic, godoc contract is honest.

PR #109, opened 2026-06-15
fix(metrics): bound network and method label cardinality 

aristidesstaffieri:
## Summary

Closes an unauthenticated availability vector in freighter-backend-v2.

`GET /api/v1/rpc-health` read an arbitrary `network` query parameter and forwarded it unchanged into Prometheus service-metric labels. Prometheus vectors keep one time series per unique label value for the life of the process — so an attacker could send requests with unique `network` values to grow memory without bound and OOM-kill the service.

This was reproduced in a 32 MiB container: exit `137`, `oom=true`.

The root cause is that `rpc-health` skipped the finite-network validation every other handler already applies. The fix validates at the handler, then adds defense-in-depth at the metrics layer so no caller-controlled value can expand label cardinality without bound.

While auditing the rest of the API for the same bug class, I found a second, lower-severity vector: the HTTP metrics middleware records the raw client-sent `method` token on every request (including unmatched 404s), and Go's server accepts arbitrary method tokens. That label is now bucketed too.

## What's in this PR

- `internal/api/handlers/rpc_health.go` — validate `network` against `{PUBLIC, TESTNET, FUTURENET}` before calling `GetHealth`; invalid values return `400` and never reach the metrics layer.
- `internal/metrics/metrics.go` — `sanitizeNetwork()` buckets any non-standard network to `"unknown"` inside `Record()`, capping the `network` label regardless of caller.
- `internal/api/middleware/metrics.go` — `sanitizeMethod()` buckets non-standard HTTP methods to `"other"`, closing the `method`-label vector on unmatched routes.
- `internal/api/handlers/rpc_health_test.go` — asserts an invalid network is rejected and the RPC service is never called.
- `internal/api/middleware/metrics_test.go` — asserts unique non-standard method tokens collapse to a single series.

## Audit notes

The only two places caller input reaches a Prometheus label are `metrics.Record` and the HTTP middleware. Both are now bounded:

- `service` / `method` (service metrics) are hardcoded constants.
- `error_type` comes from `ClassifyError` — a fixed set of kinds plus integer RPC/HTTP codes.
- `handler` is `r.Pattern` (registered route pattern, falls back to `"unknown"`); `code` is a status int.

## Test plan

- [x] `go test ./internal/...` green locally
- [x] `gofmt` clean, `go vet ./...` clean
- [ ] CI green (`build`, `check`, `Analyze (go)` currently pending)

### DIFF
--- a/internal/api/middleware/metrics.go
+++ b/internal/api/middleware/metrics.go
@@ -10,6 +10,24 @@ import (
 	"github.com/stellar/freighter-backend-v2/internal/metrics"
 )
 
+// sanitizeMethod buckets any non-standard HTTP method into a single "other"
+// label. Go's HTTP server accepts arbitrary method tokens from the client, and
+// the method flows into Prometheus metric vectors (which retain a time series
+// per unique label value). Without bucketing, a caller could send unique
+// method tokens — especially against unmatched routes, where the handler label
+// collapses to "unknown" — and grow label cardinality (and memory) without
+// bound.
+func sanitizeMethod(method string) string {
+	switch method {
+	case http.MethodGet, http.MethodHead, http.MethodPost, http.MethodPut,
+		http.MethodPatch, http.MethodDelete, http.MethodConnect,
+		http.MethodOptions, http.MethodTrace:
+		return method
+	default:
+		return "other"
+	}
+}
+
 // Metrics returns middleware that records HTTP metrics using BufferedResponseWriter.
 func Metrics(h *metrics.HTTP) Middleware {
 	return func(next http.Handler) http.Handler {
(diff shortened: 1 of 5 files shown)


internal/api/middleware/metrics.go
```diff
@@ -31,7 +49,7 @@ func Metrics(h *metrics.HTTP) Middleware {
 				handler = "unknown"
 			}
 
-			labels := []string{handler, r.Method, strconv.Itoa(code)}
+			labels := []string{handler, sanitizeMethod(r.Method), strconv.Itoa(code)}
```
aditya1702, line 52: should we expand `sanitizeHandlers` to all handlers?

aditya1702 (approved):

Issue #90, opened 2026-04-27
Implement contact endpoints (GET/PUT/DELETE /api/v1/contacts)


piyalbasu:
Implement the API endpoints specified in the design doc § API Endpoints. Server treats `contactCipher` as opaque bytes; any versioning the client needs is embedded inside the ciphertext.

See [Cross-Platform Contact Sync design doc](https://github.com/stellar/wallet-eng-monorepo/blob/main/design-docs/contact-lists/Freighter%20Authenticated%20Contact%20Sync%20Design%20Doc.md) — § API Endpoints + § Contact Storage.

**Acceptance:**
- `GET` for a new `user_id` → 200 `{"contactCipher": ""}`
- First `PUT` creates the row; subsequent `PUT`s replace
- `DELETE` removes all data for user
- Payload >256 KB → 413
- Retried `PUT` (same body) produces no duplicate state

PR #110, opened 2026-06-16
feat(db): wire up PostgreSQL — connection pool, migrations, db-health

### LINKED ISSUE #87: Wire up Postgres in freighter-backend-v2
Stand up the Go-side database layer in `freighter-backend-v2` so contact-endpoint handlers have something to talk to. Out of scope: the cluster itself (covered by the Provision Postgres issue) and the handlers themselves (covered by the Contact endpoints issue).

**Scope:**
- Add a Postgres driver and connection pool (e.g., `pgx` / `database/sql` per existing service conventions)
- Wire `DATABASE_URL` (or equivalent) via ExternalSecrets / env
- Apply the contacts table schema as a migration (whatever migration tooling the repo standardizes on)
- Extend the existing health-check endpoint to include DB connectivity
- Sane connection-pool defaults for ~6k MAU light workload

See [Cross-Platform Contact Sync design doc](https://github.com/stellar/wallet-eng-monorepo/blob/main/design-docs/contact-lists/Freighter%20Authenticated%20Contact%20Sync%20Design%20Doc.md) — Phase 1 § Postgres backend wiring; Contact Storage § Database schema.

**Acceptance:**
- `freighter-backend-v2` boots with a Postgres connection, fails fast on bad config
- Schema migration creates the `contacts` table on first run; idempotent on re-run
- Health check returns 200 only when DB is reachable; 503 otherwise
- Connection-pool config documented in repo (max conns, lifetime)

piyalbasu:
## TL;DR

Wires `freighter-backend-v2` up to PostgreSQL so future features (starting with contact lists) have a database to talk to. The service opens a connection pool on startup and fails fast if the database is misconfigured or unreachable; schema migrations run as a separate `migrate` command (never on the `serve` boot path); and a new `GET /api/v1/db-health` endpoint lets clients check whether DB-backed features are available — the same way the extension already uses `rpc-health`.

Infrastructure only — **no feature schema or endpoints** (those land in follow-up issues). Day-to-day development runs against the hosted dev/stg/prd databases (CloudNativePG); switching between a local DB and a hosted one changes exactly one thing, `DATABASE_URL`, with `make db-forward` / `make db-url ENV=<env>` helpers. Verified end-to-end against `wallet-eng-dev` (incl. through the PgBouncer pooler).

Closes #87.

<details>

**What changed:**

- **`internal/db/`** (new, mirrors `wallet-backend`): `OpenDBConnectionPool` (pgxpool, ping-to-fail-fast) + `SQLDBFromPool`; [`PoolConfig.QueryExecMode`](https://github.com/stellar/freighter-backend-v2/blob/cf5cd610ed61eaddcb9db5a6cef3192776609354/internal/db/db.go#L39-L46) set to `QueryExecModeExec` (the DB is fronted by a CNPG Pooler in **transaction** mode, where pgx's default statement caching throws `42P05`); [`Migrate`](https://github.com/stellar/freighter-backend-v2/blob/cf5cd610ed61eaddcb9db5a6cef3192776609354/internal/db/migrate.go#L17-L40) via `sql-migrate` (`ExecMaxContext`, same exec mode) over an embedded FS with an empty initial migration.
- **`cmd/migrate/`** (new): `migrate up [count]` / `down <count>` — migrations run out-of-band from `serve`, so replicas/local processes can't race a shared DB. `down` [rejects non-positive counts](https://github.com/stellar/freighter-backend-v2/blob/cf5cd610ed61eaddcb9db5a6cef3192776609354/cmd/migrate/migrate.go#L68-L73) (else `sql-migrate` rolls back everything).
- **`internal/api/serve.go`**: `initDatabase` opens the pool at boot under a bounded timeout, registers pgxpool stats as Prometheus metrics, and never migrates; pool closed on shutdown.
- **`internal/api/handlers/db_health.go`** (new): `GET /api/v1/db-health` — always `200` + `{status}` (like `rpc-health`), with a [5s singleflight cache](https://github.com/stellar/freighter-backend-v2/blob/cf5cd610ed61eaddcb9db5a6cef3192776609354/internal/api/handlers/db_health.go#L65-L92) so request volume can't exhaust the pool. `/api/v1/ping` stays dependency-free liveness.
- **`internal/metrics/pool.go`** (new): [`RegisterDBPoolMetrics`](https://github.com/stellar/freighter-backend-v2/blob/cf5cd610ed61eaddcb9db5a6cef3192776609354/internal/metrics/pool.go#L26-L31) → `freighter_db_pool_*` gauges/counters (acquired/idle/total/max conns, acquire + canceled-acquire), so pool saturation is observable.
- **`internal/config/config.go`**: `DatabaseConfig` + [`ValidatePoolConfig`](https://github.com/stellar/freighter-backend-v2/blob/cf5cd610ed61eaddcb9db5a6cef3192776609354/internal/config/config.go#L91-L106) (max≥1, min≥0, min≤max, fits int32) so bad `--db-*-conns` flags fail fast in serve.
- **`deployments/docker-compose.yml`**: pinned `postgres:16-alpine` + a one-shot `migrate` service before `api`; `DATABASE_URL` single-sourced via a YAML anchor.
- **`scripts/db-connect.sh` + `make db-forward`/`db-url`**: port-forward the CNPG primary and build `DATABASE_URL` from the basic-auth creds secret (`sslmode=require`).
- **Integration harness**: added a dedicated `freighter-postgres` container + `DATABASE_URL` so `serve` (now DB-dependent) boots in CI.
- **`internal/db/README.md`**: local + hosted connection, migrations, pooling.

**Verification:** `go build ./...`, `gofmt -l`, `go vet ./...`, full unit suite (incl. db-health cache/singleflight under `-race`) all green. Verified live against `wallet-eng-dev`: `migrate up` (idempotent), `db-health` → `healthy`, and `migrate up` **through the transaction-mode pooler** with no `42P05`. DB integration tests are gated behind `ENABLE_INTEGRATION_TESTS` (testcontainers).

- ExternalSecrets injection of `DATABASE_URL` + the per-release `migrate up` Job live in `stellar/kube` — not this PR. `serve` pings connectivity, not schema, so a deployed pod starts "healthy" against an unmigrated DB; the migrate Job must ship with any schema change.
- Generic query helpers (`QueryOne`/`QueryMany`/`RunInTransaction`) deferred to the contacts-store issue (#90).
- `wallet-eng-runbooks` reconciliation (document `db-health`) staged to apply after merge.

</details>


scripts/db-connect.sh
```diff
@@ -0,0 +1,79 @@
...
+    ;;
+  url)
+    user="$(kubectl -n "$ns" get secret "$SECRET" -o jsonpath='{.data.username}' | b64decode)"
+    pass="$(kubectl -n "$ns" get secret "$SECRET" -o jsonpath='{.data.password}' | b64decode)"
+    if [ -z "$user" ] || [ -z "$pass" ]; then
+      echo "ERROR: could not read username/password from secret ${SECRET} in ns ${ns} (is the cluster provisioned there, and is your kubectl context set to ${env}?)." >&2
+      exit 1
+    fi
+    # sslmode=require: CNPG serves TLS; we encrypt but don't verify the hostname
+    # (the cert is for the in-cluster name, not localhost).
```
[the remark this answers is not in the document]
piyalbasu: Fixed in f9a0588 — the comment now states `sslmode=require` encrypts but does **not** verify the server cert or hostname (passive-eavesdropping protection only), and points to `sslmode=verify-full` + the cluster CA (`freighter-backend-v2-db-ca`) for full verification.

internal/db/migrations/2026-06-12.0-initial.sql
```diff
@@ -0,0 +1,7 @@
+-- Initial migration. Intentionally empty: it establishes the sql-migrate
+-- tooling and the gorp_migrations bookkeeping table so the migration step runs
+-- (and is idempotent) on boot. Feature schemas are added in their own migrations.
```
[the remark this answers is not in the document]
piyalbasu: Fixed in f9a0588 — reworded to make clear migrations are applied by the `migrate up` command (a deploy Job / the docker-compose migrate service), never on the `serve` boot path.

internal/db/migrate_test.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/stellar/freighter-backend-v2/internal/db/migrations"
+)
+
+func TestMigrate_InvalidDSN(t *testing.T) {
+	t.Parallel()
+
+	// Migration on boot must fail fast on an unparseable connection string.
+	_, err := Migrate(context.Background(), "postgres://u:p@localhost:notaport/db", migrate.Up, 0)
```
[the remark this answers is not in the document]
piyalbasu: Fixed in f9a0588 — comment now reads "`migrate up` must fail fast …" instead of "on boot".

internal/config/config.go
```diff
@@ -55,6 +61,26 @@ type RedisConfig struct {
...
+
+// Validate checks required database configuration. The database is a hard
+// dependency, so an empty URL is rejected. Shared by the `serve` and `migrate`
+// commands so both fail fast with the same message.
+func (c DatabaseConfig) Validate() error {
+	if c.URL == "" {
+		return errors.New("--database-url (env DATABASE_URL) is required")
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
piyalbasu: Addressed in f9a0588 — added `DatabaseConfig.ValidatePoolConfig()` (max>=1, min>=0, min<=max, fits int32), called from serve's `PersistentPreRunE` so bad `--db-*-conns` flags fail fast with a clear message. It lives in serve rather than the shared `Validate()` because only serve sources these from flags — `migrate` opens a short-lived pool with the package defaults, so there's nothing to validate there. Covered by new config + serve tests.

cmd/serve/serve.go
```diff
@@ -37,6 +38,11 @@ func (s *ServeCmd) Command() *cobra.Command {
 			if d, m := s.Cfg.AppConfig.AccountHistoryDefaultLimit, s.Cfg.AppConfig.AccountHistoryMaxLimit; d <= 0 || m <= 0 || d > m || m > handlers.AccountHistoryUpstreamMaxLimit {
 				return fmt.Errorf("--account-history-default-limit=%d / --account-history-max-limit=%d must be positive, default <= max, and max <= %d", d, m, handlers.AccountHistoryUpstreamMaxLimit)
 			}
+			// The database is a hard dependency: fail fast with a clear message
+			// rather than surfacing an opaque connection error deeper in startup.
+			if err := s.Cfg.DatabaseConfig.Validate(); err != nil {
+				return err
```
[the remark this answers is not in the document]
piyalbasu: Fixed in 9f7767d — added a dedicated app Postgres container (`freighter-postgres`) to the integration harness and pass `DATABASE_URL` to the freighter container, so `serve` connects and `/api/v1/ping` comes up. No migration step needed for the current suite (serve only pings; no test exercises feature tables), but the DB is now in place for when #90 adds store-layer tests.

internal/db/migrate.go
```diff
@@ -0,0 +1,34 @@
...
+		return 0, fmt.Errorf("connecting to the database: %w", err)
+	}
+	defer pool.Close()
+
+	sqlDB := SQLDBFromPool(pool)
+	defer sqlDB.Close() //nolint:errcheck // best-effort close of the sql.DB wrapper
+
+	source := migrate.HttpFileSystemMigrationSource{FileSystem: http.FS(migrations.FS)}
+
+	applied, err := migrate.ExecMax(sqlDB, "postgres", source, direction, count)
```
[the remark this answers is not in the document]
piyalbasu: Fixed in 9f7767d — switched `Migrate` to `migrate.ExecMaxContext(ctx, …)` so the apply phase honors cancellation/timeout from the passed context (the CLI's `cmd.Context()`).

internal/db/migrate.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"github.com/stellar/freighter-backend-v2/internal/db/migrations"
+)
+
+// Migrate applies the embedded migrations against the database at databaseURL in
+// the given direction, up to count migrations (0 means "all"). It returns the
+// number of migrations actually applied, which is 0 on a re-run once the schema
+// is up to date — i.e. the operation is idempotent.
+func Migrate(ctx context.Context, databaseURL string, direction migrate.MigrationDirection, count int) (int, error) {
+	pool, err := OpenDBConnectionPool(ctx, databaseURL)
```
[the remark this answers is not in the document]
piyalbasu: Fixed in 1445666 — `Migrate` now opens its pool with `QueryExecModeExec` (and default sizing), matching serve. So even if the deploy migrate Job shares serve's pooler `DATABASE_URL` (CNPG Pooler, transaction mode), sql-migrate won't hit 42P05.

cmd/migrate/migrate.go
```diff
@@ -0,0 +1,101 @@
...
+			return c.Cfg.DatabaseConfig.Validate()
+		},
+	}
+
+	var configFilePath string
+	cmd.PersistentFlags().StringVar(&configFilePath, "config-path", "", "Path to config file (e.g., /etc/freighter/config.toml)")
+	cmd.PersistentFlags().StringVar(&c.Cfg.DatabaseConfig.URL, "database-url", "", "PostgreSQL connection string (env DATABASE_URL). Required.")
+
+	cmd.AddCommand(&cobra.Command{
+		Use:   "up [count]",
```
aditya1702: Why do we need a count here? Migrations are idempotent so an apply on all will still work

cmd/migrate/migrate.go
```diff
@@ -0,0 +1,101 @@
...
+			if err != nil {
+				return err
+			}
+			logger.Info("Applied migrations up", "count", applied)
+			return nil
+		},
+	})
+
+	cmd.AddCommand(&cobra.Command{
+		Use:   "down <count>",
```
aditya1702, line 59: same for the down count

cmd/migrate/migrate.go
```diff
@@ -0,0 +1,101 @@
...
+		PersistentPreRunE: func(cmd *cobra.Command, _ []string) error {
+			if err := utils.InitializeConfig(cmd); err != nil {
+				return err
+			}
+			return c.Cfg.DatabaseConfig.Validate()
+		},
+	}
+
+	var configFilePath string
+	cmd.PersistentFlags().StringVar(&configFilePath, "config-path", "", "Path to config file (e.g., /etc/freighter/config.toml)")
```
aditya1702: Why do we need a config for migration? What are we defining in this config file? We dont use any for wallet backend or horizon so we could skip it unless really needed

internal/api/handlers/db_health.go
```diff
@@ -0,0 +1,115 @@
...
+	db    DBPinger
+	ttl   time.Duration
+	now   func() time.Time
+	group singleflight.Group
+
+	mu        sync.Mutex
+	status    string
+	checkedAt time.Time
+	hasValue  bool
+}
```
aditya1702: **Do we need the caching layer here yet?**

The endpoint itself is good scaffolding (mirrors `rpc-health`), but `cachedDBHealth` (singleflight + 5s TTL) feels premature for this PR:

- **Inconsistent with `rpc-health`**, which does a live call per request with no cache. Same kind of endpoint, two different patterns.
- **No consumer yet** — nothing hits `db-health` until the DB-backed feature lands in a later PR, so there's no request volume to protect against.
- **Pool exhaustion is already observable** via the pool metrics we register, and the 2s ping timeout already bounds each probe.
- The 5s TTL adds up to 5s of stale `healthy`/`unhealthy` after the DB actually drops or recovers.

Proposal: ship with a plain `Ping` (matching `rpc-health`) and drop the cache. If load becomes a real concern once the feature lands, add caching then — ideally as a shared pattern across both health endpoints. WDYT?

internal/db/migrate_integration_test.go
```diff
@@ -0,0 +1,68 @@
+package db
```
aditya1702, line 1: nit: shouldnt this be in integrationtests/ folder?

scripts/db-connect.sh
```diff
@@ -0,0 +1,83 @@
...
+    stg) echo "wallet-eng-stg" ;;
+    prd) echo "wallet-eng-prd" ;;
+    *) return 1 ;;
+  esac
+}
+
+CLUSTER="freighter-backend-v2-db"
+SECRET="freighter-backend-v2-db-app-creds" # basic-auth: username + password
+DBNAME="freighter-backend-v2"
+RW_SVC="${CLUSTER}-rw"
```
aditya1702: **Public repo — we shouldn't hardcode our k8s topology here.** This repo is public, and the script + README bake in internal infra details:

- namespaces `wallet-eng-{dev,stg,prd}` (here + README L110/L133) — the org-wide naming convention, not just this service
- CNPG cluster `freighter-backend-v2-db`, RW svc, and the **DB-creds secret name** `freighter-backend-v2-db-app-creds` (here + README L117) — points straight at where DB creds live
- CA secret `freighter-backend-v2-db-ca` (L77) + infra state: which envs are live and the internal "Provision Postgres" work item (L23, README L110-111/L134-135)

No credentials are leaked and access still needs kubeconfig + RBAC + network, but it's free recon and there's no upside to publishing it. (wallet-backend's *public* repo only documents the local `DATABASE_URL` and keeps the hosted-from-secret steps in the private runbooks — let's match that.)

The *mechanism* (decode secret → URL-encode password → port-forward `svc/<cluster>-rw`) is standard CNPG and fine to keep public. Suggestion: stop hardcoding the *specific* names — drive them from env vars (`CNPG_CLUSTER`, `DB_CREDS_SECRET`, `DBNAME`, namespace from the current kubectl context) and document the concrete values in the private wallet-eng-runbooks. Same cleanup for the README section. WDYT?

Also I think we can just manually pass in a `DATABASE_URL` ourselves - we already do that for horizon, rpc, wallet backend and we can avoid this script.

internal/db/migrate.go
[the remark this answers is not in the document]
piyalbasu: Reverting this one. The wallet-eng team has decided to run **no** PgBouncer/CNPG Pooler — `pgxpool` already pools internally, and a transaction-mode pooler adds nothing at our current scale (double-pooling). With no pooler in front there's no 42P05 risk, so both `migrate` and `serve` now use pgx's default `CacheStatement` mode (also faster, since we connect directly to the CNPG `-rw` primary). Done in fee3f9c.

internal/db/db.go
```diff
@@ -0,0 +1,108 @@
...
+		MaxConnIdleTime: DefaultMaxConnIdleTime,
+	}
+}
+
+// resolvePoolConfig returns the first config provided, or DefaultPoolConfig() if
+// none. Callers (serve) populate all fields from flags with non-zero defaults;
+// this mirrors wallet-backend's resolvePoolConfig for cross-repo parity.
+func resolvePoolConfig(configs []PoolConfig) PoolConfig {
+	if len(configs) > 0 {
+		return configs[0]
```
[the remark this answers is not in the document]
piyalbasu: Didn't go with the blanket overlay — it would clobber *legitimate* zero values: `MinConns: 0` (no idle floor) and `QueryExecMode: 0` (pgx's default) are valid intent, not "unset", so defaulting them would be wrong.

Instead I corrected the misleading contract: `PoolConfig` is used whole-config — `serve` passes all four sizing fields, and any caller wanting a partial override starts from `DefaultPoolConfig()`. So there's no path that silently zeroes `MaxConns`. Done in fee3f9c.

cmd/migrate/migrate.go
piyalbasu: Done in fee3f9c — `up` now applies all pending migrations and takes no count (`Args: cobra.NoArgs`).

cmd/migrate/migrate.go
piyalbasu: Keeping the count on `down` (unlike `up`) — the asymmetry is deliberate. `up` is idempotent, so apply-all is the safe default. But `down` with no limit is destructive: `sql-migrate` treats `max <= 0` as "no limit", so an unbounded `down` rolls back **every** migration. The required positive count + the `< 1` guard prevent an accidental full-schema rollback. I made that reasoning explicit in the comment (fee3f9c). Happy to revisit if you'd still rather drop it.

cmd/migrate/migrate.go
piyalbasu: Done in fee3f9c — stripped it. `migrate` now takes only `DATABASE_URL` (flag or env), no `--config-path`/TOML loading, matching wallet-backend and horizon.

internal/api/handlers/db_health.go
piyalbasu: Done in fee3f9c — dropped the singleflight + 5s TTL cache. `db-health` now does a live bounded ping per request, matching rpc-health. Pool saturation stays observable via the `freighter_db_pool_*` metrics, and the 2s ping timeout still bounds each probe. If real load shows up once the DB-backed feature lands, we can add caching back as a shared pattern across both health endpoints.

internal/db/migrate_integration_test.go
piyalbasu: Kept it in `internal/db` for now — it's a focused white-box integration test of `db.Migrate` (testcontainers, env-gated), as opposed to the full-service container tests in `internal/integrationtests/` that spin up the whole app. Happy to move it if you'd rather all container-based tests live in one place.

scripts/db-connect.sh
piyalbasu: Done in fee3f9c — removed `scripts/db-connect.sh` and the `make db-forward`/`db-url` targets entirely, and stripped the hardcoded namespaces / cluster / creds-secret names from the README. Hosted connection now follows the same manual-`DATABASE_URL` approach as horizon/RPC/wallet-backend; the concrete per-env values move to the private wallet-eng-runbooks. Only the generic CNPG mechanism stays (port-forward `svc/<cluster>-rw` + build the URL), with placeholders.

cmd/migrate/migrate_test.go
```diff
@@ -0,0 +1,64 @@
...
+func TestMigrateCmd_RejectsEmptyDatabaseURL(t *testing.T) {
+	t.Parallel()
+
+	migrateCmd := &MigrateCmd{Cfg: &config.Config{}}
+	cmd := migrateCmd.Command()
+	cmd.SetOut(io.Discard)
+	cmd.SetErr(io.Discard)
+	// `migrate up` with no --database-url must fail fast in PersistentPreRunE,
+	// before any attempt to touch a database.
+	cmd.SetArgs([]string{"up"})
```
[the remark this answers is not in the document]
piyalbasu: Good catch — confirmed and fixed in 5df125a. With `DATABASE_URL` exported (which the new README recommends for local work), the env fallback in `PersistentPreRunE` let this test run `migrate up` against the real DB instead of exercising the missing-config path — I reproduced it failing at the pool-ping stage. The test now clears `DATABASE_URL` via `t.Setenv` (and drops `t.Parallel`, which `t.Setenv` forbids), so it's hermetic and can't touch a live database.

cmd/migrate/migrate.go
aditya1702: ah makes sense, 👍

aditya1702 (approved):

cmd/serve/serve_test.go
```diff
@@ -86,6 +86,22 @@ func TestServeCmd_Execute(t *testing.T) {
...
+func TestServeCmd_RejectsEmptyDatabaseURL(t *testing.T) {
+	t.Parallel()
+
+	serveCmd := &ServeCmd{Cfg: &config.Config{}}
+	cmd := serveCmd.Command()
+	cmd.RunE = func(*cobra.Command, []string) error { return nil }
+	cmd.SetOut(io.Discard)
+	cmd.SetErr(io.Discard)
+	// No --database-url provided: the DB is a hard dependency, so boot must fail fast.
+	cmd.SetArgs([]string{})
```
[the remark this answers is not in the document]
piyalbasu: Fixed in cf9e1f1 — same hardening as the migrate missing-url test. `TestServeCmd_RejectsEmptyDatabaseURL` now calls `t.Setenv("DATABASE_URL", "")` (and drops `t.Parallel`, which `t.Setenv` forbids), so `utils.InitializeConfig` can't bind a shell-exported `DATABASE_URL` into `--database-url` and mask the missing-config path. Verified it still passes with `DATABASE_URL` exported.

internal/db/migrate_integration_test.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	ctx := context.Background()
+	container, err := postgres.Run(ctx,
+		"postgres:16-alpine",
+		postgres.WithDatabase("freighter"),
+		postgres.WithUsername("freighter"),
+		postgres.WithPassword("freighter"),
+		testcontainers.WithWaitStrategy(
+			wait.ForListeningPort("5432/tcp").WithStartupTimeout(60*time.Second),
```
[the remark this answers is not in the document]
piyalbasu: Fixed in 32aa3d0. Switched the test container to `postgres.BasicWaitStrategies()`, which waits for the "ready to accept connections" log twice (the official image starts, runs init, then restarts) and then the port — so the first `Migrate`/`Ping` can't race the restart.

Also applied the same fix to the app Postgres container in the integration harness (`createAppPostgresContainer`), which had the identical port-only wait and sits in `serve`'s boot-ping path — mirrored via `wait.ForAll(ForLog(...).WithOccurrence(2), ForListeningPort(...))`.

Issue #88, opened 2026-04-27
Implement JWT auth middleware


piyalbasu:
Port `wallet-backend`'s `JWTManager` pattern to `freighter-backend-v2`. Verify Ed25519 signature against `sub` (= the auth pubkey), validate expiry, `bodyHash`, and `methodAndPath` claims.

See [Cross-Platform Contact Sync design doc](https://github.com/stellar/wallet-eng-monorepo/blob/main/design-docs/contact-lists/Freighter%20Authenticated%20Contact%20Sync%20Design%20Doc.md) — Auth Flow + JWT claims table.

**Acceptance:**
- Valid JWT → 200
- Expired / tampered / wrong-key → 401

PR #112, opened 2026-06-23
feat(auth): stateless Ed25519 JWT auth middleware (#88)

piyalbasu:
## TL;DR

Adds stateless JWT authentication to `freighter-backend-v2`, closing #88. A client proves its identity by signing a short-lived token per request with a key derived from the user's seed; the server verifies that signature on its own, with no accounts, sessions, database lookups, or server-side secrets to manage.

This PR ships **only the auth building block plus a tiny demo endpoint** — it does **not** turn on authentication for any existing endpoint, so current Freighter clients are completely unaffected. Enforcement is staged behind a config switch (`AUTH_MODE`, default "permissive") so we can require tokens later, once clients have rolled out, by flipping one setting.

There's one new endpoint, `GET /api/v1/auth/whoami`, which simply tells a caller which user the server thinks they are — useful for client teams validating their token construction.

<details>

**Design doc:** [`docs/design/2026-06-22-jwt-auth-middleware.md`](https://github.com/stellar/freighter-backend-v2/blob/9401710cafd1058ea6f0a17bca49a9b70c3c74c7/docs/design/2026-06-22-jwt-auth-middleware.md)

**Identity model (diverges from wallet-backend deliberately).** Modeled on wallet-backend's `JWTManager`, but where wallet-backend verifies against a *known/allowlisted* server key, here the JWT's `sub` claim **is** the caller's unified user ID — a hex-encoded Ed25519 public key derived from the user's seed — which doubles as the signature-verification key. So there is no server key/secret to provision and no allowlist — the token proves "whoever signed this controls the private key for user ID `<sub>`". `sub` is hex (not a strkey `G...` address), per the design.

**What changed, by area:**
- [`internal/auth/`](https://github.com/stellar/freighter-backend-v2/blob/9401710cafd1058ea6f0a17bca49a9b70c3c74c7/internal/auth/parser.go) — the verifier primitive: `Claims` + `Validate` (lifetime ≤ 15s, `methodAndPath` + `bodyHash` binding), `ParseJWT` (pins `alg=EdDSA` via `WithValidMethods` to block alg-confusion; ±5s `WithLeeway` for mobile clock skew; derives the key from `sub`), `HTTPRequestVerifier`, typed errors, request-context helpers, and `Mode`.
- [`internal/api/middleware/auth.go`](https://github.com/stellar/freighter-backend-v2/blob/9401710cafd1058ea6f0a17bca49a9b70c3c74c7/internal/api/middleware/auth.go) — `Auth(verifier, mode, metrics)`. Truth table: **permissive** (no token → anonymous; present-but-invalid → 401) vs **required** (missing/invalid → 401). On success the user ID is attached to the request context.
- Config — `AUTH_MODE` / `--auth-mode` (default `permissive`), validated at startup and resolved to a typed `auth.Mode` once in `ApiServer.Start()`.
- Metrics — `freighter_auth_requests_total{result,reason}` for rollout adoption signal; `reason` is a bounded taxonomy (`expired` / `bad_signature` / `bad_timing` / `bad_method_path` / `bad_body_hash` / `bad_subject` / `malformed`) so rejection spikes can be triaged by cause. Each rejection also logs `reason` + `detail` (never token/body bytes).
- [`internal/api/handlers/whoami.go`](https://github.com/stellar/freighter-backend-v2/blob/9401710cafd1058ea6f0a17bca49a9b70c3c74c7/internal/api/handlers/whoami.go) + wiring in `serve.go` — the only route wrapped with `Auth`; returns the standard `{data:{...}}` envelope.
- `go.mod` — promotes `golang-jwt/jwt/v5` from indirect to direct.

**Rollout semantics.** All gated endpoints will share one mode and flip together (client adoption is per-app-version, not per-endpoint), so the mode is a single global config value rather than a per-route argument. A present-but-invalid token is rejected (401) in *both* modes, which keeps #88's acceptance criteria true throughout the rollout and gives a clean adoption signal.

**Verification:** `go build ./...`, `go vet ./...`, `gofmt -l` clean, and the full `go test ./...` suite green. Acceptance (#88: valid → 200, expired/tampered/wrong-key → 401) is covered by the middleware truth-table test and an end-to-end `whoami` route test; the auth package has table-driven unit tests including alg-confusion rejection, leeway boundaries, body/query binding, and a >256 KB body (no truncation). A code review (8 finder angles + verification) was run and its findings addressed.

- No feature endpoints or existing-endpoint gating (separate tickets). When a follow-up sets `AUTH_MODE=strict` and gates real endpoints, it should add a runbook in `wallet-eng-runbooks` for the `freighter_auth_requests_total{result="rejected"}` cutover signal (modeled on the existing `ONRAMP_AUTH_MODE` runbook).
- `whoami` is intentionally minimal and could be dropped if reviewers prefer zero new production routes; acceptance is also covered by middleware tests.
- `HashBody`/`ExpiredTokenError` overlap wallet-backend's `auth` package; not deduped because the real signing client is the TS extension/mobile, not wallet-backend, so importing wouldn't provide protocol-safety.

</details>


internal/api/serve.go
```diff
@@ -145,6 +155,14 @@ func (s *ApiServer) initHandlers() (*http.ServeMux, error) {
 	}
 	mux.HandleFunc("GET /api/v1/accounts/{address}/transactions", handlers.CustomHandler(accountHistoryHandler.GetAccountTransactions))
 
+	// Auth: a guarded endpoint that echoes the authenticated user ID. It exercises
+	// the JWT middleware end-to-end and gives client teams a surface to validate
+	// their JWT construction. s.authMode is resolved once in Start().
+	verifier := auth.NewVerifier()
+	whoamiHandler := handlers.NewWhoamiHandler()
+	mux.Handle("GET /api/v1/auth/whoami",
```
piyalbasu, line 163: this /whomai endpoint can be removed once we're actually using the middleware in practice

internal/logger/logger.go
```diff
@@ -60,19 +60,31 @@ func New(cfg loggerConfig) *Logger {
...
 func Global() *Logger {
-	once.Do(func() {
-		if global == nil {
-			global = New(DefaultConfig())
-		}
-	})
+	mu.Lock()
+	defer mu.Unlock()
+	if global == nil {
+		global = New(DefaultConfig())
```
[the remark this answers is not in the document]
piyalbasu: Fixed in 07a2043. `Global()` now takes a read lock on the hot path (every Debug/Info/Warn/Error goes through it) and only escalates to the write lock for first-time init. `SetOutput()` still takes the write lock. Verified clean under `go test -race`.

internal/api/middleware/auth.go
```diff
@@ -0,0 +1,66 @@
...
+			case err == nil:
+				metrics.RecordAuth(authMetrics, "authenticated", "ok")
+				r = r.WithContext(auth.ContextWithUserID(r.Context(), userID))
+
+			case errors.Is(err, auth.ErrNoToken):
+				if mode == auth.Required {
+					metrics.RecordAuth(authMetrics, "rejected", "no_token")
+					httperror.Unauthorized("", nil).Render(w)
+					return
+				}
```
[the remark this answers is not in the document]
piyalbasu: Fixed in 07a2043 — returns `"unauthorized"` now. Confirmed every other httperror call site in the codebase passes a non-empty message, so this aligns them.

internal/api/middleware/auth.go
```diff
@@ -0,0 +1,66 @@
...
+				// category and request method/path, never the token or body bytes.
+				reason := auth.Reason(err)
+				metrics.RecordAuth(authMetrics, "rejected", reason)
+				logger.InfoWithContext(r.Context(), "rejected request with invalid auth token",
+					"reason", reason,
+					"detail", err.Error(),
+					"method", r.Method,
+					"path", r.URL.Path)
+				httperror.Unauthorized("", nil).Render(w)
+				return
```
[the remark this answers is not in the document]
piyalbasu: Fixed in 07a2043 — `"unauthorized"`. The detailed reason stays in the structured log line only; the client body remains generic.

internal/api/middleware/auth.go
```diff
@@ -0,0 +1,66 @@
...
+
+			default:
+				// Operational failure (e.g. reading the body).
+				metrics.RecordAuth(authMetrics, "rejected", "internal")
+				logger.ErrorWithContext(r.Context(), "auth check failed", "error", err)
+				httperror.InternalServerError("", err).Render(w)
+				return
```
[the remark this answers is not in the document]
piyalbasu: Fixed in 07a2043 — uses `"An unexpected error occurred"`, matching the Recover middleware. The internal error detail is still server-side only (`Err` is `json:"-"`).

docs/design/2026-06-22-jwt-auth-middleware.md
```diff
@@ -0,0 +1,192 @@
...
+risk silently truncating the bytes it hashes.
+
+`ParseJWT(tokenString, methodAndPath, body)`:
+1. `ParseUnverified` to read `claims.Subject`.
+2. `claims.Validate(methodAndPath, body, MaxTokenLifetime)`:
+   - `exp` and `iat` set; `exp - iat ≤ MaxTokenLifetime`;
+   - `methodAndPath` matches `"<METHOD> <RequestURI>"` (binds query string);
+   - `bodyHash == HashBody(body)`;
+   - `Subject` is valid hex decoding to exactly `ed25519.PublicKeySize` (32) bytes.
+3. Decode `Subject` → `ed25519.PublicKey`.
```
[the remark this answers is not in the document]
piyalbasu: Fixed the doc in 07a2043: moved the Subject hex/length validation out of the `Claims.Validate` bullet list and into step 3, noting it happens in `ParseJWT` after `Validate`, not inside `Validate`. Kept the code boundary as-is.

internal/auth/verifier.go
```diff
@@ -0,0 +1,73 @@
...
+	return &Verifier{}
+}
+
+// VerifyHTTPRequest extracts the bearer token, binds it to the request's
+// method+path and body, and verifies it. It returns ErrNoToken when no bearer
+// token is present (so callers can allow anonymous access in permissive mode);
+// any other error indicates a present-but-invalid token.
+func (v *Verifier) VerifyHTTPRequest(r *http.Request) (string, error) {
+	// RFC 6750: the "Bearer" auth scheme is case-insensitive.
+	scheme, token, found := strings.Cut(r.Header.Get("Authorization"), " ")
```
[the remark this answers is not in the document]
piyalbasu: Fixed in 07a2043 — added `Authorization` to `Access-Control-Allow-Headers` in `internal/api/middleware/header.go` so the preflight does not block JWT-bearing requests once clients send them.

internal/auth/verifier.go
```diff
@@ -0,0 +1,73 @@
...
+func readAndResetBody(r *http.Request) ([]byte, error) {
+	if r.Body == nil {
+		return nil, nil
+	}
+	b, err := io.ReadAll(r.Body)
+	if err != nil {
+		return nil, err
+	}
+	r.Body = io.NopCloser(bytes.NewReader(b))
+	return b, nil
```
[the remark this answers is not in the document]
piyalbasu: Not making this change — I do not think it is a leak for the standard `net/http` server. The server captures its own reference to the original body when it builds the `*response` (`response.reqBody = req.Body`, set in `conn.readRequest` before the handler runs) and closes *that* in `finishRequest()` after `ServeHTTP` returns. It does not close whatever the handler has since assigned to `r.Body`. So swapping `r.Body` for the `NopCloser` does not orphan the original — the original is still closed via the server's saved `reqBody`. Adding an explicit `Close()` after `io.ReadAll` would be harmless but redundant, so I am leaving it out. Happy to revisit if this verifier is ever driven by a non-net/http server that does not retain that reference.

internal/api/middleware/auth.go
```diff
@@ -0,0 +1,66 @@
...
+					"method", r.Method,
+					"path", r.URL.Path)
+				httperror.Unauthorized("unauthorized", nil).Render(w)
+				return
+
+			default:
+				// Operational failure (e.g. reading the body).
+				metrics.RecordAuth(authMetrics, "rejected", "internal")
+				logger.ErrorWithContext(r.Context(), "auth check failed", "error", err)
+				httperror.InternalServerError("An unexpected error occurred", err).Render(w)
```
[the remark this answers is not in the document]
piyalbasu: Fixed in c0b0390. The middleware now detects the oversized-body case and renders a 413 (`RequestEntityTooLarge`, "Request body too large"), matching the body-reading handlers, instead of falling through to the 500 default. `IsMaxBytesError` was switched to `errors.As` so it matches through the verifier's `%w`-wrapped error, and the outcome is recorded under a new `too_large` auth metric reason. Added `TestAuth_OversizedBodyReturns413` covering the path end-to-end.

internal/auth/parser.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		return nil, &VerificationError{Reason: ReasonBadSubject, Err: err}
+	}
+
+	_, err = jwtgo.ParseWithClaims(tokenString, claims,
+		func(*jwtgo.Token) (any, error) { return pubKey, nil },
+		// Pin the algorithm to EdDSA to prevent alg-confusion (e.g. alg=none or
+		// an HS256 token forged with the public key as the HMAC secret).
+		jwtgo.WithValidMethods([]string{"EdDSA"}),
+		jwtgo.WithLeeway(ClockSkewLeeway),
```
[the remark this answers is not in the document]
piyalbasu: Fixed in 70f0c4b. `Claims.Validate` now rejects `iat` more than `ClockSkewLeeway` in the future (`ReasonBadTiming`), so a token cannot be dated ahead of now beyond the intended ±skew. The exp and iat future bounds are now checked against a single captured `now`. Added `TestParseJWT_FutureIssuedAt` (your `iat=exp=now+20s` example is rejected) and documented the bounds in the design doc.

internal/auth/claims.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+const (
+	// MaxTokenLifetime caps how long a token may be valid (exp - iat). Short
+	// lifetimes are the primary replay defense.
+	MaxTokenLifetime = 15 * time.Second
+	// ClockSkewLeeway tolerates clock drift between client and server when
+	// checking expiry, per the design doc (mobile clients especially).
+	ClockSkewLeeway = 5 * time.Second
+)
```
JakeUrban, line 19: Should we make these configurable?

internal/auth/claims.go
```diff
@@ -0,0 +1,74 @@
...
+		return &VerificationError{Reason: ReasonBadTiming, Err: errors.New("missing iat claim")}
+	}
+
+	lifetime := c.ExpiresAt.Sub(c.IssuedAt.Time)
+	if lifetime < 0 {
+		return &VerificationError{Reason: ReasonBadTiming, Err: errors.New("exp precedes iat")}
+	}
+	if lifetime > maxLifetime {
+		return &VerificationError{Reason: ReasonBadTiming, Err: fmt.Errorf("token lifetime %s exceeds max %s", lifetime, maxLifetime)}
+	}
```
JakeUrban, line 51: I think generally we want to minimize the information given to unauthorized clients, so maybe we should remove the maximum lifetime, as well as the mention of a clock skew leeway below.

internal/auth/context.go
```diff
@@ -0,0 +1,20 @@
+package auth
+
+import "context"
+
+type contextKey struct{}
+
+var userIDKey = contextKey{}
```
JakeUrban: Is this idiomatic?

internal/auth/errors.go
```diff
@@ -0,0 +1,78 @@
...
+// middleware renders any error matching it as a 401.
+var ErrUnauthorized = errors.New("not authorized")
+
+// ExpiredTokenError marks a token that failed verification solely because it
+// expired (beyond the clock-skew leeway). It wraps ErrUnauthorized so callers
+// that don't care about the distinction still treat it as a 401.
+type ExpiredTokenError struct {
+	ExpiredBy time.Duration
+	Err       error
+}
```
JakeUrban, line 25: Maybe `ErrExpiredToken` to be consistent with the other errors?

internal/auth/parser.go
```diff
@@ -0,0 +1,72 @@
...
+func decodePublicKey(sub string) (ed25519.PublicKey, error) {
+	raw, err := hex.DecodeString(sub)
+	if err != nil {
+		return nil, fmt.Errorf("subject is not valid hex: %w", err)
+	}
+	if len(raw) != ed25519.PublicKeySize {
+		return nil, fmt.Errorf("subject decodes to %d bytes, want %d", len(raw), ed25519.PublicKeySize)
+	}
+	return ed25519.PublicKey(raw), nil
+}
```
JakeUrban, line 78: What is the bytes aren't a valid ed25519 key? Does `ed25519.PublicKey()` not return an error?

JakeUrban (approved):
LGTM with some comments.

One question I had was whether we plan to check whether the JWT's subject is a known freighter user, in addition to validating the JWT signature and claims, in a future PR? This check would require the set of our users' IDs in a DB table, which itself implies users are registered with the backend likely during the user's onboarding flow.

I think this DB table is needed anyway once we start storing other data in association to the users.

piyalbasu:
> LGTM with some comments.
> 
> One question I had was whether we plan to check whether the JWT's subject is a known freighter user, in addition to validating the JWT signature and claims, in a future PR? This check would require the set of our users' IDs in a DB table, which itself implies users are registered with the backend likely during the user's onboarding flow.
> 
> I think this DB table is needed anyway once we start storing other data in association to the users.

This is something we could do in the future, but as you noted, would need users to have already been seeded in our DB. Are you viewing this from a security angle, .i.e, denying users access if they have not been registered? That is possible to do, but we would need to retroactively register all existing users first

internal/auth/claims.go
piyalbasu: It's a good thought, but I think we should err on the side of not doing this because changing these values affects client token generation: tightening `MaxTokenLifetime` would require us to make sure the client tokens to also tighten their lifetime. Because the client and server need to (roughly) move in sync, it's probably best to have these hardcoded

internal/auth/claims.go
piyalbasu: Good idea - we should just log this rather than exposing

internal/auth/context.go
piyalbasu: Yes, this pattern is idiomatic for a single key (vs using a string, for ex). https://pkg.go.dev/context#WithValue

Actually, one improvement we can make here is to used a named struct to future-proof this against multiple keys

internal/auth/errors.go
piyalbasu: This is actually an error type, analogous to `VerificationError`, which carry more data than just the error. The `Err_` naming convention is used for sentinels which are just the error string. This is the pattern we use in the rest of this repo

internal/auth/parser.go
piyalbasu: Good question - interestingly: `ed25519.PublicKey()` is just a type-conversion. I was surprised by this, too. 

It can't return an error or do any kind of validation. The length check above is the actual guard. Non-valid keys that are the correct length safely fail at the signature validation level

JakeUrban: Lets make a comment calling that out, just so its clear that returned keys from this function aren't necessarily valid ed25519 public keys

JakeUrban:
> Are you viewing this from a security angle, .i.e, denying users access if they have not been registered?

That is how I was thinking about it, although we're really only adding friction for an attacker, rather than preventing anything, since the attacker could just figure out how to register an ID before proceeding.

> but we would need to retroactively register all existing users first

We will need to figure out how to retroactively (and moving forward, during onboarding) register users in a database on our backend eventually, since its required once we store more data associated with those users. But that can be out of scope of this PR, and we can bundle it in with contact lists, assuming thats the first data we'll store on our backend in association with users.

The only other data I'd like to have persisted is users' public keys for BI-reasons. Maybe this is something we can persist in amplitude as a user-attribute though.

piyalbasu:
> > but we would need to retroactively register all existing users first
> 
> We will need to figure out how to retroactively (and moving forward, during onboarding) register users in a database on our backend eventually, since its required once we store more data associated with those users. But that can be out of scope of this PR, and we can bundle it in with contact lists, assuming thats the first data we'll store on our backend in association with users.
> 


Yeah we can definitely retroactively register users: either through contact lists or maybe a more used feature like sending/swapping or signing.

> The only other data I'd like to have persisted is users' public keys for BI-reasons. Maybe this is something we can persist in amplitude as a user-attribute though.

Yeah, it might be useful to store public keys/user id's in Amplitude. It'd likely be easier to query and build charts there

internal/auth/verifier.go
```diff
@@ -0,0 +1,73 @@
...
+		// Operational failure, not an auth failure: don't wrap ErrUnauthorized.
+		return "", fmt.Errorf("reading request body: %w", err)
+	}
+
+	methodAndPath := fmt.Sprintf("%s %s", r.Method, r.URL.RequestURI())
+	claims, err := ParseJWT(token, methodAndPath, body)
+	if err != nil {
+		return "", err
+	}
+	return claims.Subject, nil
```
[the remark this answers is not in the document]
piyalbasu: Fixed in f052248. `ParseJWT` now canonicalizes the subject before returning it: after the signature verifies, it sets `claims.Subject = hex.EncodeToString(pubKey)`, so any upper/mixed-case `sub` collapses to the canonical lowercase form. The decoded bytes are identical either way, so the signature still verifies — only the string handed to `auth.UserIDFromContext` changes, which keeps storage/cache keying from splitting one signing key into two users.

https://github.com/stellar/freighter-backend-v2/blob/f0522487cdd8d1c1e71bed000fec9123be464938/internal/auth/parser.go#L59-L63

Added `TestParseJWT_CanonicalizesSubject` (uppercase `sub` → lowercase user ID).

internal/auth/verifier.go
```diff
@@ -0,0 +1,73 @@
...
+// any other error indicates a present-but-invalid token.
+func (v *Verifier) VerifyHTTPRequest(r *http.Request) (string, error) {
+	// RFC 6750: the "Bearer" auth scheme is case-insensitive.
+	scheme, token, found := strings.Cut(r.Header.Get("Authorization"), " ")
+	if !found || !strings.EqualFold(scheme, "Bearer") {
+		return "", ErrNoToken
+	}
+	token = strings.TrimSpace(token)
+	if token == "" {
+		return "", ErrNoToken
```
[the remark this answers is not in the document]
piyalbasu: Fixed in f052248. A `Bearer` scheme with an empty or whitespace-only credential — and a bare `Bearer` with nothing after it — now returns `VerificationError{Reason: ReasonMalformed}` (which wraps `ErrUnauthorized`), so it's rejected with 401 in **both** modes instead of being classified as `ErrNoToken` and passed through as anonymous in permissive. A genuinely absent or non-Bearer `Authorization` header still returns `ErrNoToken`, so true anonymous requests are unaffected and the rollout truth table stays honest.

https://github.com/stellar/freighter-backend-v2/blob/f0522487cdd8d1c1e71bed000fec9123be464938/internal/auth/verifier.go#L44-L52

Added `TestVerifyHTTPRequest_EmptyBearerRejected` covering `Bearer `, `Bearer`, whitespace-only, and lowercase `bearer ` → all 401 with reason `malformed`.

PR #116, opened 2026-07-01
fix(prices): return null for any zero price instead of "0"

aristidesstaffieri:
## Summary

The `/api/v1/token-prices` endpoint is documented to return `null` for any unpriceable token, but a token that Stellar Expert *knows* yet has no price for (illiquid, no `price` field in the 200 response) came back as `currentPrice: "0"` instead. That "0" is a Go zero-value leak, not a real price — in the extension it reads as a present value, suppresses the `null`-gated spot-price fallback, and shows a misleading $0.00.

Root cause: an absent `price` field decodes to the Go zero value `0`, indistinguishable from a genuine upstream price of 0. Rather than try to tell them apart, this PR treats any zero price as unpriceable — a real 0 is just as unpriceable as an absent one — and honors the documented `null` contract.

## What's in this PR

- `internal/types/interfaces.go` — `StellarExpertAsset.Price` stays a plain `float64`; comment documents that any `0` (absent field or genuine) means unpriceable.
- `internal/services/prices.go` — `fetchFromUpstream` gates on `asset.Price == 0`, resolving to `(nil, true)` so it serializes as `null` and caches as an authoritative miss (same as not-found/malformed).
- `internal/services/prices_test.go` — merges the two zero cases into one `TestPrices_ZeroPrice_ReturnsNull`, drops the `floatPtr` helper, updates fixtures to plain `float64`.
- `internal/services/stellar_expert_test.go` — asserts on the now-value `Price` field.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/services/... ./internal/api/handlers/...` pass
- [x] New case verified: `TestPrices_ZeroPrice_ReturnsNull`
- [x] Confirmed live against Stellar Expert: `USDyc:GDDRGUTO...` returns 200 with `price` absent (fixed → `null`); `WSGBP:GDSVWEA7...` returns a real price
- [ ] CI green

### DIFF
--- a/internal/services/prices.go
+++ b/internal/services/prices.go
@@ -291,11 +291,12 @@ func (p *pricesService) fetchFromUpstream(ctx context.Context, network, cacheNet
 		return nil, false
 	}
 
-	// A known asset can return 200 with no `price` field (illiquid, no recent
-	// trades). That is unpriceable, not a price of 0: honor the documented
-	// null contract rather than leaking a zero value. Resolves to (nil, true)
-	// like not-found/malformed so it caches as an authoritative miss.
-	if asset.Price == nil {
+	// A zero price — whether Stellar Expert omitted the `price` field (a known
+	// but illiquid asset; JSON absence decodes to 0) or reported a genuine 0 —
+	// is unpriceable. Honor the documented null contract rather than leaking a
+	// "0" string. Resolves to (nil, true) like not-found/malformed so it caches
+	// as an authoritative miss.
+	if asset.Price == 0 {
 		return nil, true
 	}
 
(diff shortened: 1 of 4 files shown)


internal/services/prices.go
```diff
@@ -291,6 +291,14 @@ func (p *pricesService) fetchFromUpstream(ctx context.Context, network, cacheNet
 		return nil, false
 	}
 
+	// A known asset can return 200 with no `price` field (illiquid, no recent
+	// trades). That is unpriceable, not a price of 0: honor the documented
+	// null contract rather than leaking a zero value. Resolves to (nil, true)
+	// like not-found/malformed so it caches as an authoritative miss.
+	if asset.Price == nil {
```
aditya1702: According to this condition, stellar expert will return null? The PR description says that it returns 0 right?

aristidesstaffieri: this is now a `*float64` so the nil check is on the pointer

aristidesstaffieri: @aditya1702 fa5b18f45861d347516945628a3ba4234d3bc014 resolves this for the "0" case also

aditya1702 (approved):

PR #130, opened 2026-07-09
fix(balances): derive is_funded from native balance, hide balances when unfunded

aditya1702:
## Problem

`GetBalancesByAccountAddresses` set `is_funded=true` on any successful wallet-backend fetch, equating "wallet-backend has data for this address" with "the account is funded". That's wrong for:
- **merged accounts** — history but no classic account,
- **never-created addresses**,
- **contract-token-only holders** — hold Soroban tokens, no classic account.

(An earlier attempt fixed this in wallet-backend by returning `accountByAddress: null` for such accounts, but that also nulled their transactions/state-changes — every account-scoped query gates on `accountByAddress == nil` — breaking the account view for merged accounts. So the fix belongs here, where the wrong assumption lives.)

## Fix

Derive `is_funded` from the presence of a `NativeBalance` in the returned balances — the classic account exists iff it has a native balance — and hoist `subentry_count` from that same entry. An unfunded account (no native balance) is reported with an empty balance set: any residual Soroban token balances it holds are not surfaced.

wallet-backend still serves the account's transactions/state-changes for merged accounts; this change only governs the balances envelope.

## Effect

- funded account → `is_funded=true`, balances shown (unchanged).
- merged / never-created / contract-token-only → `is_funded=false`, empty balances (was wrongly `is_funded=true`).

## Testing

- New `unfunded_account_with_only_contract_balance_hides_balances`: a successful fetch with only a SEP-41 balance → `is_funded=false`, `subentry_count=0`, empty balances.
- Existing envelope/fan-out tests unchanged. `go test -race ./...`, whole-repo tests, and `make lint` (0 issues) are green.

### DIFF
--- a/internal/services/wallet_backend.go
+++ b/internal/services/wallet_backend.go
@@ -191,7 +191,8 @@ func (w *walletBackendService) GetBalancesByAccountAddresses(ctx context.Context
 			// even if a future SDK regression returned nil for an account with zero
 			// balances. GetAllAccountBalances currently returns a non-nil empty
 			// slice; keeping the guard is cheap insurance. IsFunded defaults to
-			// false and is set true only on a successful fetch (below).
+			// false and is set true only when the fetched balances include a
+			// native balance (below).
 			ab := &types.AccountBalances{
 				Address:  addr,
 				Balances: []types.Balance{},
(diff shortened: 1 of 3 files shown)


internal/services/wallet_backend.go
```diff
@@ -215,22 +216,30 @@ func (w *walletBackendService) GetBalancesByAccountAddresses(ctx context.Context
...
-				// success, not from the absence of an error, and stays correct even
-				// if the error handling above changes.
-				ab.IsFunded = true
-				mapped := make([]types.Balance, 0, len(balances))
+				// "Funded" means the classic account exists, which on-chain is exactly
+				// the presence of a native balance. A successful fetch only proves
+				// wallet-backend has data for the address — which also holds for merged
+				// accounts (history but no classic account) and holders of only Soroban
+				// tokens. So derive IsFunded from the native balance rather than from the
+				// fetch succeeding, hoisting SubentryCount from the same entry.
```
aristidesstaffieri, line 226: Doc drift: the function doc above ("The single address-scoped outcome is the typed wbclient.ErrAccountNotFound sentinel…", lines 155–159) — and the mirrored doc on `types.AccountBalances` (`internal/types/wallet_backend.go:21–26`) — still claim account-not-found is the *only* address-scoped path to `is_funded=false`. This change adds a second one: a successful fetch whose balances contain no native entry. Both docs should describe the native-balance gate.

internal/services/wallet_backend_test.go
```diff
@@ -627,6 +627,36 @@ func TestGetBalancesByAccountAddresses_EnvelopeEnrichment(t *testing.T) {
 		assert.EqualValues(t, 0, results[0].SubentryCount)
 		assert.Empty(t, results[0].Balances)
 	})
+
+	t.Run("unfunded_account_with_only_contract_balance_hides_balances", func(t *testing.T) {
```
aristidesstaffieri, line 679: Coverage gap: the funded-path tests all use a single native balance, so nothing asserts the new `if ab.IsFunded` block still surfaces non-native balances when the account *is* funded. A mixed native+SEP-41 fixture asserting both entries appear would catch a regression inside the gate (e.g. accidentally filtering to native-only) that currently passes the whole suite.

internal/services/wallet_backend.go
```diff
@@ -215,22 +216,30 @@ func (w *walletBackendService) GetBalancesByAccountAddresses(ctx context.Context
...
 				for _, b := range balances {
```
aristidesstaffieri, line 228: Nit: the native-detection loop and the mapping loop can fold into one pass — append `mapBalance(b)` while detecting the native entry, then `if ab.IsFunded { ab.Balances = mapped }`. One traversal, and future balance-derived fields have only one loop to touch.

internal/services/wallet_backend.go
aditya1702: Good catch — fixed in 1a64e23. Updated both docs (the `GetBalancesByAccountAddresses` godoc and `types.AccountBalances`) to state that `is_funded=false` now has two address-scoped paths: the `ErrAccountNotFound` sentinel *and* a successful fetch whose balances contain no native entry (merged / contract-token-only accounts). Resolved.

internal/services/wallet_backend_test.go
aditya1702: Agreed — fixed in 1a64e23. Added `funded_account_surfaces_native_and_non_native_balances`: a funded native + SEP-41 fixture asserting both balances are surfaced (and `is_funded`/`subentry_count` come from the native entry). It fails if the funded mapping is ever narrowed to native-only. Resolved.

internal/services/wallet_backend.go
aditya1702: Done in 1a64e23. Folded into a single pass — map each balance while detecting the native entry, then `if ab.IsFunded { ab.Balances = mapped }`. One traversal, one place for future balance-derived fields. Resolved.

aristidesstaffieri (approved):

Issue #106, opened 2026-06-07
Instrument auth + contacts endpoints (metrics + structured logging)


piyalbasu:
Add the Prometheus metrics and structured logging the design doc requires for the auth + contacts surface. These are the basis for confirming the endpoints work on prod (success rate, latency, 401 rate) and for the Phase 1 alerts — so they're part of the definition of done for #88 and #90, not a later add-on.

See [Cross-Platform Contact Sync design doc](https://github.com/stellar/wallet-eng-monorepo/blob/main/design-docs/contact-lists/Freighter%20Authenticated%20Contact%20Sync%20Design%20Doc.md) — § Observability.

**Metrics (Prometheus /metrics):**
- `freighter_auth_jwt_validations_total{result}` — success / expired / invalid_signature / body_mismatch / method_mismatch
- `freighter_contacts_operations_total{operation, status}` — get/put/delete × success/error
- `freighter_request_duration_seconds{method, path}` — latency histogram

**Structured logging (slog via internal/logger):**
- Every authenticated request: user_id, method, path, status_code, latency_ms, iss
- JWT validation failures: reason (token_expired / invalid_signature / body_hash_mismatch / method_path_mismatch)
- NFR: never log the auth pubkey in a way that correlates to a G address; no PII.

**Acceptance:**
- All three metrics emit on /metrics with correct labels under real traffic.
- Auth-failure logs carry the reason label.
- Verifiable on prod: auth success rate and contacts op success rate can be computed; p95 latency visible.

Relates to #56 / #63 (generic metrics) but is scoped to the auth/contacts surface.

piyalbasu:
**Implementation note — reuse existing metrics infra (`internal/metrics/metrics.go`); don't stand up new latency/request metrics.**

The service already exposes `GET /metrics` (`internal/api/serve.go:112`, promhttp over a custom registry) and has a metrics package with HTTP + Service groups. Mapping this ticket's metrics against what already exists:

| Design-doc metric | Status | Action |
| --- | --- | --- |
[...]

PR #131, opened 2026-07-13
feat(auth): apply auth middleware to all user-facing routes (#114)

### LINKED ISSUE #114: Apply auth middleware to all endpoints (permissive global; required on user-scoped)
Apply the `auth.Auth` middleware across all `/api/v1` routes in **permissive** mode, instead of only `GET /api/v1/auth/whoami`.

Today only `whoami` is wrapped (`internal/api/serve.go`); the other endpoints (`/protocols`, `/collectibles`, `/ledger-key/accounts`, `/feature-flags`, `/accounts/balances`, `/token-prices`, `/accounts/{address}/transactions`, `/rpc-health`) are unauthenticated.

- Wrap all routes with `middleware.Auth(verifier, mode, …)` in **permissive** mode: a valid JWT populates the request context (`auth.UserIDFromContext`) and auth metrics; absent/anonymous requests still succeed.
- Reserve **required** mode for user-scoped routes (contacts, when they land) — out of scope here.

Part of the Unified User Model auth rollout (option A: gate only authenticated-context routes; keep anonymous routes open). Permissive-global is safe to ship ahead of clients — it never rejects anonymous traffic.

**Acceptance:**
- Every `/api/v1` endpoint runs the auth middleware.
- A request with a valid JWT populates user context + auth metrics on every endpoint.
- Anonymous requests still succeed on all currently-open endpoints (no client breakage).

piyalbasu:
## TL;DR

Closes #114. Every user-facing `/api/v1` endpoint now runs the JWT auth middleware, not just `whoami`. Behavior is unchanged by default (`permissive`: anonymous requests still succeed, valid tokens populate user context + auth metrics, present-but-invalid tokens are rejected). Flipping the single `--auth-mode strict` switch now requires a valid JWT on **all** user-facing routes at once — the intended end state once clients send JWTs. The infra health probes (`/ping`, `/db-health`, `/rpc-health`) are deliberately left open in every mode so they can't be locked out.

**Deviation from the ticket:** the issue body lists `/rpc-health` as a route to gate, but this PR exempts it alongside the other health probes for operational safety (see below). Flagging for reviewer sign-off.

**Related:** runbook reconciliation for the broader strict-mode 401 surface — [stellar/wallet-eng-runbooks#20](https://github.com/stellar/wallet-eng-runbooks/pull/20).

<details>

**What changed:**

- [`internal/api/serve.go`](https://github.com/stellar/freighter-backend-v2/blob/f5abbcd90ade6353cf1da8015da6f13427c345e9/internal/api/serve.go#L178) — `initHandlers()` now binds one shared middleware value `authed := middleware.Auth(verifier, s.authMode, s.appMetrics.Auth)` and wraps each of the 8 user-facing routes with it (`protocols`, `collectibles`, `ledger-key/accounts`, `feature-flags`, `accounts/balances`, `token-prices`, `accounts/{address}/transactions`, `auth/whoami`). The three health probes stay bare `mux.HandleFunc(...)`. whoami's previous bespoke inline `Auth` wrap is folded into the shared value. One shared value = flipping `--auth-mode` moves all wrapped routes together.
- **No changes** to `internal/auth` (verifier, `context.go`), `middleware.Auth`, or the global middleware chain in `initMiddleware`. This is route registration only.
- [`docs/design/2026-06-22-jwt-auth-middleware.md`](https://github.com/stellar/freighter-backend-v2/blob/f5abbcd90ade6353cf1da8015da6f13427c345e9/docs/design/2026-06-22-jwt-auth-middleware.md) rewritten to describe the current state (all user-facing routes gated, health bare), dropping the original #88 in/out-of-scope framing.

**Why per-route (not a global chain middleware):** wrapping per route keeps `Auth` *inside* the mux, so it runs **after** `http.ServeMux` has stamped `r.Pattern`. The outer `Metrics` middleware therefore still reads the real route label for authenticated requests — a global `Auth` above the mux would fork the request via `WithContext` before routing and collapse authenticated traffic to `handler="unknown"`. A regression test guards this.

**`/rpc-health` deviation:** gating it under `strict` risks 401ing a k8s/synthetic probe (pod churn), and `db-health`/`ping`/`rpc-health` are health surfaces that must answer without a JWT. Exempted deliberately. **Follow-up:** confirm in `stellar/kube` whether anything probes `/rpc-health`; if not, gating it later is a one-line change.

**Verification:**
- `go test ./internal/api/...` — all pass. New/changed tests: user-facing route respects auth mode (permissive anonymous→200, permissive invalid-token→401, strict anonymous→401); health probes anonymous in strict; valid token populates whoami (`authenticated:true` + user ID); authenticated request keeps its real `freighter_http_requests_total` handler label (the `r.Pattern` regression guard); and a route-set guard asserting **every** user-facing route 401s anonymously in strict (catches a future route added without `authed(...)`).
- `go build ./...`, `go vet ./internal/api/...`, `gofmt -l` — clean.
- Pre-existing unrelated `internal/services` test failure (`TestNewRPCService_GetLedgerEntry/returns_error_on_network_failure`, an IPv6-vs-IPv4 `localhost` exact-string assertion) fails identically on `main` and is untouched by this branch.

- Runbook reconciliation for the broader strict-mode 401 surface is staged separately in `wallet-eng-runbooks` (4xx-ratio + protocols-probe runbooks).
- Per-request auth logging/metrics enrichment (`user_id`, `iss`) and the mutable auth-context holder that requires are #106, intentionally not here.

</details>

### DIFF
--- a/internal/api/serve.go
+++ b/internal/api/serve.go
@@ -178,18 +178,19 @@ func (s *ApiServer) closeServices() {
 func (s *ApiServer) initHandlers() (*http.ServeMux, error) {
 	mux := http.NewServeMux()
 
-	// Initialize health check handler (dependency-free liveness signal)
+	// Health/liveness/readiness probes: registered BARE — never wrapped by Auth.
+	// K8s and the docker-compose wget healthcheck cannot present per-request JWTs,
+	// and db-health is designed never to fail the request; gating any of these
+	// would 401 probes under `--auth-mode strict` and cause pod churn.
 	healthHandler := handlers.NewHealthHandler()
 	mux.HandleFunc("GET /api/v1/ping", handlers.CustomHandler(healthHandler.CheckHealth))
 
-	// Initialize RPC health check handler
 	rpcHealthHandler := handlers.NewRPCHealthHandler(s.rpcService)
 	mux.HandleFunc("GET /api/v1/rpc-health", handlers.CustomHandler(rpcHealthHandler.CheckRPCHealth))
 
-	// Initialize DB health check handler (reports connectivity in the body; never
-	// fails the request, so it can't restart or depool a pod over a DB outage).
-	// Pass a true nil interface when disabled — a nil *pgxpool.Pool boxed in the
-	// interface is not == nil, which would defeat the handler's disabled check.
+	// Pass a true nil interface when the DB is disabled — a nil *pgxpool.Pool
+	// boxed in the interface is not == nil, which would defeat the handler's
+	// disabled check.
 	var dbPinger handlers.DBPinger
 	if s.dbPool != nil {
 		dbPinger = s.dbPool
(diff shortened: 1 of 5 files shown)


internal/api/serve.go
```diff
@@ -178,42 +178,50 @@ func (s *ApiServer) closeServices() {
...
+	// Health/liveness/readiness probes: registered BARE — never wrapped by Auth.
+	// K8s and the docker-compose wget healthcheck cannot present per-request JWTs,
+	// and db-health is designed never to fail the request; gating any of these
+	// would 401 probes under `--auth-mode strict` and cause pod churn.
 	healthHandler := handlers.NewHealthHandler()
 	mux.HandleFunc("GET /api/v1/ping", handlers.CustomHandler(healthHandler.CheckHealth))
 
-	// Initialize RPC health check handler
 	rpcHealthHandler := handlers.NewRPCHealthHandler(s.rpcService)
 	mux.HandleFunc("GET /api/v1/rpc-health", handlers.CustomHandler(rpcHealthHandler.CheckRPCHealth))
```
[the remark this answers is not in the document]
piyalbasu: Good catch, and agreed this shouldn't be closed silently — going with option (a): the `/rpc-health` exemption is **deliberate**, not an oversight.

Rationale for not gating it (option b): `/ping`, `/db-health`, and `/rpc-health` are infra health surfaces hit by k8s liveness/readiness probes (and a docker-compose `wget` healthcheck on `/ping`), none of which can present a per-request Ed25519 JWT. Under `--auth-mode strict` a gated probe would 401 and drive pod churn — a worse failure than an anonymous connectivity check. `db-health` in particular is designed to *never* fail the request, so gating it would defeat its purpose. "JWT-capable probing" for a liveness check isn't practical here.

This is documented in the PR description (the "Deviation from the ticket" note) and in the design doc's *Deviation from the ticket body* section, with a follow-up to confirm in `stellar/kube` whether anything actually probes `/rpc-health` (if nothing does, gating it later is a one-line change).

Action on (a): I'll reconcile #114's acceptance criteria to record the three health probes as exempt so closing the ticket is accurate — rather than gate `/rpc-health`.

internal/api/serve.go
```diff
@@ -223,15 +220,53 @@ func (s *ApiServer) initHandlers() (*http.ServeMux, error) {
 	if err != nil {
 		return nil, fmt.Errorf("init account-history handler: %w", err)
 	}
-	mux.HandleFunc("GET /api/v1/accounts/{address}/transactions", handlers.CustomHandler(accountHistoryHandler.GetAccountTransactions))
+	whoamiHandler := handlers.NewWhoamiHandler()
+
+	return []route{
+		// Health/liveness/readiness probes: gated=false, registered BARE — never
```
piyalbasu, line 226: When we start migrating v1 endpoints to v2, we can easily separate "authed" endpoints from non-auth endpoints in this array

internal/api/serve.go
```diff
@@ -223,15 +220,53 @@ func (s *ApiServer) initHandlers() (*http.ServeMux, error) {
...
+		// wrapped by Auth. K8s and the docker-compose wget healthcheck cannot present
+		// per-request JWTs, and db-health is designed never to fail the request;
+		// gating any of these would 401 probes under `--auth-mode strict` and cause
```
piyalbasu, line 229: to be clear, we are still in "permissive" mode - meaning endpoints that don't pass a JWT will still work

JiahuiWho (approved):

PR #132, opened 2026-07-13
test(auth): guard all user-facing routes stay gated (#114 follow-up)

piyalbasu:
## TL;DR

Optional hardening follow-up to #131: a single test that enumerates **every** user-facing `/api/v1` route and asserts each one rejects an anonymous request in `strict` mode. It exists to catch a future route being added without the shared auth wrapper — a silent "fail-open" where an endpoint ships unauthenticated by accident. Test-only, no production change.

**How involved:** barely — **+39 lines, one table test, no stub, no production change.** It turned out much lighter than feared (see below).

> Stacked on #131 (base = `feat/apply-auth-all-endpoints`). Merge or retarget to `main` after #131 lands.

<details>

**What changed:** one test appended to `internal/api/serve_test.go` — `TestApiServer_initHandlers_AllUserFacingRoutesGatedInStrict`. It builds the mux with `AuthMode: "strict"` and, for each of the 8 user-facing routes, sends an anonymous request and asserts `401`.

**Why it's cheap (the key insight):** in `strict` mode the `Auth` middleware rejects an anonymous request with 401 **before** the wrapped handler runs. So the test never reaches the handlers — the nil `rpcService`/`walletBackendService`/`pricesService` left by the `newTestAPIServer` helper are never invoked, and **no service stub is needed** (unlike the health-probe test in #131, which does invoke `/rpc-health` and therefore needed a stub). That's why this is ~35 lines instead of a stub-heavy fixture.

**What it catches:** a route registered bare (`mux.HandleFunc` without `authed(...)`) would, in strict, reach its handler instead of 401ing — returning 200/404 or panicking on a nil service — and fail the assertion. This is the fail-open guard the final review flagged as optional.

**Scope boundary:** the complementary direction (health probes must stay anonymous in strict) is already covered by `TestApiServer_initHandlers_HealthRoutesAnonymousInStrict` in #131; this test only covers the gated set.

**Verification:** `go test ./internal/api/...` passes; `gofmt -l` clean.

</details>

### DIFF
--- a/internal/api/serve_test.go
+++ b/internal/api/serve_test.go
@@ -379,3 +379,42 @@ func TestApiServer_authenticatedRequestKeepsRouteMetricLabel(t *testing.T) {
 	unknown := testutil.ToFloat64(s.appMetrics.HTTP.RequestsTotal.WithLabelValues("unknown", "GET", "200"))
 	assert.Equal(t, float64(0), unknown, "authed request must not collapse to the unknown handler label")
 }
+
+// TestApiServer_initHandlers_AllUserFacingRoutesGatedInStrict guards the full set
+// of user-facing routes against one being added (or left) without the shared
+// authed wrapper — a silent fail-open. In strict mode the Auth middleware rejects
+// an anonymous request with 401 BEFORE the handler runs, so this needs no service
+// stubs: the nil services left by newTestAPIServer are never reached. A route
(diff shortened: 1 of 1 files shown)


piyalbasu:
Folded into #131 (cherry-picked as commit 9d07419). Closing this stacked PR; the guard test now lives on the #114 branch.

PR #139, opened 2026-07-21
feat(config): add Blend config (cache TTLs, earn-pools allowlist path)

JiahuiWho:
#### Description:

First of ~4 PRs adding a Blend positions API to freighter-backend-v2, backed by wallet-backend's Blend GraphQL surface (stellar/wallet-backend#661 stack).Planned follow-ups: 
(2) GraphQL decode types + client methods on WalletBackendService, 
(3) the per-address positions endpoint, 
(4) the pools/earn-options catalog endpoints.

This PR adds configuration only, no behavior change, nothing consumes the values yet:
  - `BlendConfig` with three settings:
    - `--blend-positions-cache-ttl-seconds` (default 30): Redis TTL for per-address position responses. User-visible staleness is this TTL plus wallet-backend's ingestion lag, so it stays short.
    - `--blend-catalog-cache-ttl-seconds` (default 60): Redis TTL for the address-independent market views (pools, earn options); one cache entry per network serves every user.
    - `--earn-pools-config-path` (default empty): path to a JSON allowlist of pool contract IDs offered in the Earn flow. Curates earn-options only, user positions are never filtered, since users may hold positions in non-curated pools. Empty disables curation.
  - Startup validation rejecting negative TTLs, matching the existing PersistentPreRunE checks.


internal/config/config.go
```diff
@@ -146,3 +147,23 @@ type WalletBackendConfig struct {
...
 }
+
+// BlendConfig tunes the Blend positions/catalog endpoints, which are served
+// from wallet-backend's Blend GraphQL surface (URLs and signing keys come
+// from WalletBackendConfig).
+type BlendConfig struct {
+	// PositionsCacheTTLSeconds is the Redis TTL for per-address position
+	// responses. User-visible staleness is this TTL plus wallet-backend's
+	// own ingestion lag, so keep it short.
+	PositionsCacheTTLSeconds int
```
aditya1702: I dont think we should cache the positions here - I know they would not change a lot between simultaneous requests, however we should start with just calling WB on each position request. Similar to the balances queries. The caching makes sense for the pool information that is always fixed

internal/config/config.go
```diff
@@ -146,3 +147,23 @@ type WalletBackendConfig struct {
...
+	// CatalogCacheTTLSeconds is the Redis TTL for the address-independent
+	// market views (pools, earn options). One cache entry per network serves
+	// every user.
+	CatalogCacheTTLSeconds int
+	// EarnPoolsConfigPath points to a JSON allowlist of pool contract IDs
+	// per network that Freighter offers for new deposits. It curates the
+	// earn-options endpoint only — user positions are never filtered by it,
+	// since users may hold positions in non-curated pools. Empty path
+	// disables curation (all pools are offered).
+	EarnPoolsConfigPath string
```
aditya1702, line 164: What does this config store exactly? I dont see it defined in the other 3 downstream PRs

internal/config/config.go
JiahuiWho: Agreed. I'll remove the positions TTL here and the caching from the positions service in PR 3; the catalog cache stays since that's one shared  entry per network.

internal/config/config.go
JiahuiWho: It's the earn-flow curation allowlist, a JSON of pool contract addresses per network (`{"PUBLIC": [...], "TESTNET": [...]}`) that Freighter offers for new deposits. Consumed in PR 4 to filter the earn-options.

Will add blend fixedv2 pool on mainnet and testnet.

aditya1702 (approved):

PR #140, opened 2026-07-22
feat(services): add Blend GraphQL client (positions, pools, earn option)

JiahuiWho:
Second of ~4 PRs adding a Blend positions API to freighter-backend-v2, backed by wallet-backend's Blend GraphQL surface (stellar/wallet-backend#661 stack). Follow-ups: 
(3) the per-address positions endpoint, 
(4) the pools/earn-options catalog endpoints.

This PR adds the wallet-backend client layer
  - **Three new `WalletBackendService` methods** wrapping the wbclient SDK's typed Blend API (`GetBlendPools`, `GetAccountBlendPositions` from stellar/wallet-backend#664):
    - `GetBlendPositions`: an account unknown to the indexer (`ErrAccountNotFound`) normalizes to empty positions 
    - `GetBlendPools`: the pool-wide catalog; always a non-nil slice.
    - Errors classify through the existing `classifyWBError`, so these calls behave exactly like the balances/history calls.
  - No freighter-side GraphQL or decode types: query documents and types
    live in `pkg/wbclient`, next to the schema they mirror.

 **Dependencies**: pins wallet-backend to the `blend/pr6-integration-tests` branch tip for development. Will re-pin to `main` (one-line bump) once the wallet-backend Blend stack lands there.

  **Testing**
 unit tests with fake servers covering the wrapper policies; an opt-in live smoke test (`WB_LIVE_URL`) for verifying against a real instance; validated end-to-end against the wallet-backend testnet dev instance earlier in development.


JakeUrban:
Should we make the `wbclient` changes for the Blend API before merging this PR? It should reduce the code in this one significant since the client should have the type definitions right?

aditya1702 (approved):
@JiahuiWho LGTM. nit: update the PR description to reflect the direct client calls instead of the graphql queries.

Also we will have to update the go.mod, go.sum entries to point to `main` once the wallet backend BLEND PRs are merged.

PR #141, opened 2026-07-22
feat(api): add GET /accounts/{address}/positions

JiahuiWho:
Maps wallet-backend Blend positions into the frontend-shaped response: one payload serves Position Home (header + pool rows) and Position Details (per-asset supply/borrow rows).

- Mapper: fully-exited (all-zero) reserve rows are filtered from display lists (upstream emits them for earnings history); supply rows combine plain-supply and collateral portions while preserving the split; `interest_earned_usd` converts the token-denominated lifetime interest at today's oracle price (null when price/decimals unavailable, never a fabricated zero).
- Account header: total = sum of pool net values with strict null propagation; net APY = net-value-weighted mean of pool net APYs. Both rules are isolated in accountAggregate() pending upstream confirmation.
- Redis cache per (network, address), TTL from BlendConfig (default 30s); cache reads fail open, cache errors never fail the request.
- Handler follows house patterns: network + strkey validation, 10s timeout, translateServiceError mapping (upstream errors -> 502), empty


internal/services/positions.go
```diff
@@ -0,0 +1,234 @@
...
+			logger.ErrorWithContext(ctx, "positions cache write failed", "error", cacheErr)
+		}
+	}
+	return result, nil
+}
+
+// mapAccountPositions shapes the upstream Blend positions into the response.
+func mapAccountPositions(upstream *wbtypes.BlendAccountPositions) *types.AccountPositions {
+	positions := make([]types.PoolPosition, 0, len(upstream.Pools))
+	for _, pool := range upstream.Pools {
```
aditya1702, line 85: Note that while we dont support initiating backstops right now, we do want to render current backstop positions on the UI. So we should iterate through both the pools and backstop

internal/services/positions.go
```diff
@@ -0,0 +1,234 @@
...
+// wallet-backend's own ingestion lag.
+func (p *positionsService) GetAccountPositions(ctx context.Context, address, network string) (_ *types.AccountPositions, err error) {
+	start := time.Now()
+	defer func() {
+		metrics.Record(p.svcMetrics, positionsServiceName, "GetAccountPositions", network, time.Since(start).Seconds(), err)
+	}()
+
+	cacheKey := fmt.Sprintf("%s:%s:%s", positionsCacheKeyPrefix, strings.ToLower(network), address)
+	if p.redis != nil {
+		hits, cacheErr := p.redis.MGetJSON(ctx, []string{cacheKey}, func() any { return &types.AccountPositions{} })
```
aditya1702: Lets update this to remove the redis cache for positions

internal/services/positions.go
```diff
@@ -0,0 +1,234 @@
...
+		svcMetrics:    m,
+	}
+}
+
+func (p *positionsService) Name() string { return positionsServiceName }
+
+// GetAccountPositions returns the account's positions, cached per
+// (network, address) for cacheTTL. User-visible staleness is the TTL plus
+// wallet-backend's own ingestion lag.
+func (p *positionsService) GetAccountPositions(ctx context.Context, address, network string) (_ *types.AccountPositions, err error) {
```
aditya1702: As per our discussion, lets make this endpoint multi-account with parallel fan-out

internal/services/positions.go
JiahuiWho: Dropped in 20f7f91703cd90050611ce6b45ea7882b7e33888

internal/services/positions.go
JiahuiWho: Fixed in 5726316f56089ea3b5d5a33abe01474767ddafde

internal/services/positions.go
JiahuiWho: Fixed in c65d9c07536df326e62c4f55daacde7aff458ac8

aditya1702 (approved):

PR #142, opened 2026-07-22
feat(api): add Blend market catalog endpoints (pools, earn options)

JiahuiWho:
Last of 4 PRs adding the Blend positions API (config → GraphQL client →positions endpoint → market endpoints). 

  | Route | Screens |
  |---|---|
  | `GET /api/v1/protocols/blend/pools?network=` | Pool Details (market view) |
  | `GET /api/v1/protocols/blend/earn-options?network=` | Earn: select token → select pool |
  
  - **Earn options are derived in this repo from the pools catalog**: one entry per asset with at least one enabled reserve in a pool whose status accepts deposits, filtered through the operator allowlist (`--earn-pools-config-path`, JSON per network). Curation applies to the earn flow only, the pools catalog and user positions are never filtered. A configured-but-unreadable allowlist fails startup instead of silently offering every pool.
  - One response serves both earn screens: pick a token from `options[]`, then a pool from that option's `pools[]` (sorted by supplied USD descending, unpriced pools last; headline rate = `supply_apy + emissions_supply_apr`).
  - Earn options read through the pools catalog's cache, so both endpoints share one upstream query and one Redis cache entry per network (60s default); the derivation itself runs per request.

  **Testing**: allowlist loading (missing/malformed/case-insensitivity), earn-options derivation (status-based pool filtering, disabled-reserve exclusion, allowlist curation, per-network scoping, supplied-USD ordering with unpriced tie-break), catalog mapping incl. not-yet-ingested pools (all-null), and handler validation/error/empty cases. Verified live against the wallet-backend dev instance: 57 pools unfiltered, and a one-pool allowlist narrowing earn assets to that pool's reserves.


internal/services/blend_catalog.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+	return out
+}
+
+// deriveEarnOptions groups supply-eligible reserves asset-first: pools whose
+// status accepts deposits (per the SDK's status table; null status means the
+// pool's config is not yet ingested and is excluded), reserves that are
+// enabled, and — when an allowlist is configured — pools Freighter curates.
+// Assets are ordered by asset id; each asset's pools by supplied USD
+// descending (unpriced last, id tie-break), mirroring the ordering the
```
aditya1702: Please remove comments referencing the old design - we want to keep only fresh comments. Make this change for all comments in this design anywhere else too

aditya1702 (approved):
LGTM

Issue #147, opened 2026-07-31
Wall clock skew causes degraded experience


aristidesstaffieri:
### What version are you using?
latest

### What did you do?
Our tolerance for a time expiration on the v2 jwts are currently 15s. Under normal circumstances, some systems can experience wall clock skew well into the minutes range. When a user is on a system experiencing this, they are locked out of any Freighter experience that relies on the v2 APIs.

### What did you expect to see?
Better tolerance for wall clock skew, or a different age out mechanism

### What did you see instead?
401s from users with wall clock skew.
Some auth libraries have converged on 5 minutes as the gate time, but we can never be sure that we completely fix this for any instance of time skew. Is there a different way to gate the jwts that does not require clients and servers to agree on wall time?

JakeUrban:
Noting that the backend has a leeway of 2 minutes by default, and it is configurable. @piyalbasu noted in his implementation that we'd need to monitor expired JWT errors to dial in the appropriate leeway.

I think we either:

- Create a view in grafana / loki that shows us the frequency of the log that indicates an expired JWT was rejected, and we update the leeway over time
- Remove the `exp` claim from the JWT. Obviously we'd need to access the risks of doing so prior.

aristidesstaffieri:
> Noting that the backend has a leeway of 2 minutes by default, and it is configurable. [@piyalbasu](https://github.com/piyalbasu) noted in his implementation that we'd need to monitor expired JWT errors to dial in the appropriate leeway.
> 
> I think we either:
> 
> * Create a view in grafana / loki that shows us the frequency of the log that indicates an expired JWT was rejected, and we update the leeway over time
> * Remove the `exp` claim from the JWT. Obviously we'd need to access the risks of doing so prior.

I think the hard part about this approach is that there is no relation between one user's clock skew and another's. We could set a limit according to what we see in traffic and still risk new users having longer skew. 

One potentially way to harden this would be to use the server's time in the payload to build the jwt in this failure case. The 401 rejection carries payload has a Date header so in the 401 case, clients can retry with a new jwt built on that time + offset.

piyalbasu:
I pulled the prod numbers for the last 24h. Two things worth knowing before we pick a leeway value: the tolerance is already 2m rather than 15s, and **no leeway value fixes this** — the clocks involved are wrong by hours, not seconds.

[...]

PR #148, opened 2026-08-01
fix(auth): serve clock-skewed tokens anonymously in permissive mode (#133)

### LINKED ISSUE #133: Permissive auth: pass through invalid JWTs (fail-open but observable) instead of 401
## Summary

During the JWT auth rollout, `--auth-mode` defaults to `permissive` so we don't break clients that don't yet send JWTs. But today a **present-but-invalid** JWT is rejected with `401` in *both* modes. That means a client that attaches a *malformed* JWT to a previously-open endpoint (e.g. `GET /api/v1/protocols`) gets a `401` on a call that would have succeeded anonymously — the act of (buggily) adopting auth breaks the request, which undercuts the point of permissive mode.

## Proposal: fail-open but observable (permissive only)

In **permissive** mode, when a token is present but invalid (`auth.ErrUnauthorized`), keep the existing metric + structured log, then **pass the request through as anonymous** (no `userID` in context) instead of returning `401`. Users keep working; we still get a clear internal signal that clients are sending broken JWTs.

The middleware already emits both signals in the reject branch (`internal/api/middleware/auth.go`) — the only change is to record + log and then fall through to `next.ServeHTTP` in permissive, rather than rendering `401`:

```go
case errors.Is(err, auth.ErrUnauthorized):
    metrics.RecordAuth(authMetrics, <label>, reason)                          // keep
    logger.InfoWithContext(r.Context(), "rejected request with invalid auth token",
        "reason", reason, "method", r.Method, "path", r.URL.Path)             // keep
    if mode == auth.Required {
        httperror.Unauthorized("unauthorized", nil).Render(w)
        return
    }
    // permissive: fall through as anonymous (no userID attached)
```

## Scope / behavior

- **Permissive only.** `strict` stays fail-closed: present-but-invalid → `401`.
- **Narrow to `ErrUnauthorized`** (bad signature / expired / bad body-hash / bad method-path / malformed). Keep `too_large` → `413` and operational errors → `500` unchanged — those aren't "client sent a slightly-wrong JWT."
- **Distinct metric label** (e.g. `result="invalid_permitted"` alongside today's `authenticated` / `anonymous` / `rejected`) so the invalid-token rate stays measurable and can gate the eventual permissive→strict flip.
- **`whoami`** with a bad token in permissive degrades to `{"authenticated": false}` (safe; the client sent a token but is treated as anonymous). Future user-scoped / `Required` routes (e.g. contacts) are strict and unaffected.

## Acceptance criteria

- In `permissive`: a present-but-invalid JWT on a user-facing route → request succeeds (normal anonymous result), and a metric + a log line record the invalid token with its `reason`.
- In `strict`: a present-but-invalid JWT → `401` (unchanged).
- `too_large` → `413` and operational errors → `500` in both modes (unchanged).
- The invalid-in-permissive outcome is distinguishable in metrics from both `authenticated` and anonymous `no_token`.
- Tests cover: permissive invalid-token passes through (+ metric/log emitted), strict invalid-token → `401`, and whoami with a bad token in permissive → `authenticated: false`.

## Notes

- This **reverses** the behavior documented for #88 ("permissive present-but-invalid → 401") and in the JWT auth design doc's mode table — both should be updated as part of this change.
- Follows #114 / #131 (auth applied to all user-facing routes). Motivation: keep permissive faithful to "don't break clients during rollout" while still surfacing client JWT-construction bugs internally, rather than turning a client bug into user-facing 401s.

piyalbasu:
## TL;DR

Users whose device clock is badly wrong are currently locked out of every Freighter feature that talks to this service. Their requests are correctly signed and would be served happily if they sent no credentials at all — but because they *do* send one, and it carries a timestamp we disagree with, we reject them. That defeats the purpose of the permissive rollout mode, which exists precisely so the auth migration doesn't break anyone.

This makes permissive mode treat a wrong clock — in either direction, running fast or slow — as a failure to authenticate rather than a reason to refuse service: those requests are now served anonymously, exactly as if no credentials had been presented. A malformed or mis-bound credential is still rejected, and strict mode is unchanged.

Over the 7 days ending 2026-08-03 this accounted for **520 of the 524 rejections — 99.2%**, each affected device failing 100% of the time. Widening the existing clock-tolerance setting cannot fix it: that setting is hard-capped at 10m, and the largest value it permits still leaves well over half of these users locked out. The clocks in question are wrong by hours to weeks, not seconds.

Both directions matter, and this is the part that changed since the first draft: fast clocks were 0 in the initial 24h sample, but are now 14.9% of rejections and **the majority on the two most recent days**. A fix covering only slow clocks would already be leaving a growing population behind.

Closes #133. Mitigates #147 (does not close it — see below).

<details>

**What changed**

- [`internal/api/middleware/auth.go`](https://github.com/stellar/freighter-backend-v2/blob/dda760f/internal/api/middleware/auth.go) — the `ErrUnauthorized` branch computes `permitted := mode == auth.Permissive && (reason == auth.ReasonExpired || reason == auth.ReasonClockAhead)`. When true it records, logs, then `break`s out of the switch to the existing `next.ServeHTTP` — the same anonymous path a no-token request already takes, with no `userID` attached. Everything else still renders 401 and returns.
- [`internal/auth/errors.go`](https://github.com/stellar/freighter-backend-v2/blob/dda760f/internal/auth/errors.go) / [`claims.go`](https://github.com/stellar/freighter-backend-v2/blob/dda760f/internal/auth/claims.go) — new `ReasonClockAhead` and `ClockAheadError{AheadBy}`; the two future-bound checks in `Claims.Validate` now return it instead of `ReasonBadTiming`.
- [`internal/metrics/metrics.go`](https://github.com/stellar/freighter-backend-v2/blob/dda760f/internal/metrics/metrics.go) — new exported `ResultInvalidPermitted = "invalid_permitted"` result label.
- Log line keeps `reason`/`detail`/`method`/`path`, gains a `permitted` field, and its message changes from `"rejected request with invalid auth token"` to `"invalid auth token"` (the old wording would be false for served requests). Verified nothing in `wallet-eng-runbooks` greps either the old message or the metric.
- Design doc mode table updated, including why the original rationale was wrong.

**Which reasons are permitted, and why the timing bucket got split**

`expired` and `clock_ahead` — the two directions a clock can be wrong. `clock_ahead` is new in this PR, split out of `bad_timing`.

That split is the substance of the change. `bad_timing` was never a clock signal: it bundled two genuine clock checks (`iat` or `exp` too far ahead) with three malformed-token checks (missing `exp`/`iat`, `exp` preceding `iat`, over-long lifetime). Permitting the whole bucket would serve malformed tokens and make `invalid_permitted` — the counter that gates the strict flip — mix wrong clocks with client bugs. Rejecting the whole bucket, which an earlier revision of this PR did, leaves fast-clock users locked out for a reason unrelated to them. So neither side of the bucket was the right unit; the bucket was.

New `ReasonClockAhead` + `ClockAheadError`, the latter carrying `AheadBy` so the overshoot is diagnosable from logs exactly as `ExpiredTokenError.ExpiredBy` already is for lagging clocks. `bad_timing` keeps the malformed cases and still 401s.

**One consequence worth naming explicitly**

`clock_ahead` is assigned in `Claims.Validate`, which runs *before* signature verification — so a **forged** token dated into the future is now served, without its signature ever being checked.

That is safe rather than merely tolerated: permitting attaches no `userID`, so the request is byte-for-byte equivalent to one carrying no `Authorization` header, which these routes already serve anonymously. An attacker gains nothing they could not have by sending no token at all. `TestAuth_PermissiveForgedTokenOutcomes` pins all three forged-token outcomes and asserts the invariant that makes it safe — a forged token never yields a `userID`.

The cost is measurement, not access: read `invalid_permitted{reason="clock_ahead"}` as an upper bound on fast-clock clients rather than an exact count. `expired` has no such caveat (jwt/v5 returns on signature failure before validating claims, so it implies a verified signature).

**Why this grants nothing**

No `userID` enters the request context on this path, so no downstream authorization decision can consume the token. `whoami` is the only handler in the service that reads `auth.UserIDFromContext`, so for the four endpoints actually affected in prd (`protocols`, `token-prices`, `collectibles`, `ledger-key/accounts`) the response is identical authenticated or not.

**Why leeway-widening was rejected**

**The decisive reason is structural, not numerical:** `--auth-clock-skew-leeway` is validated at startup to `>= 0 && <= 10m` (`cmd/serve/serve.go:51`). No legal value of the flag covers this population, so leeway-widening is not a candidate fix regardless of which value you'd pick. Raising the cap is what a leeway fix would actually require, and full coverage of the observed offsets needs a ~19.8-day leeway — a ~40-day replay window (`2·leeway + 15s`).

The rejections resolve into a small number of fixed clock offsets rather than a drift distribution — each stable to within a second across hundreds of samples, and stable per device across days. Over the 7 days ending 2026-08-03 the lagging-clock offsets were 3h01m20s (50.7%), ~3m04s (44.3%), 19h06m, 8h00m00s (exact — a timezone written as absolute time), 475h14m (~19.8 days), plus a handful of one-offs.

Coverage by leeway value, measured against the 442 `expired` rejections in that window:

| leeway | accepts | still 401 |
| --- | --- | --- |
| 2m (current default) | 0.0% | 442 |
| 5m | 34.8% | 288 |
| **10m (the hard cap)** | **44.3%** | **246** |
| 3h05m | 95.9% | 18 |
| 480h | 100% | 0 |

An earlier revision of this description reported `5m 0/335` and `20m 107/335`. Those were wrong: `ExpiredTokenError.ExpiredBy` is `time.Since(claims.ExpiresAt)`, i.e. measured against `exp` itself and not against `exp + leeway`, so client lag is `expired_by + 15s` and the ~3m04s cluster does fall under a 5m leeway. The conclusion is unchanged — and the hard cap makes it hold without depending on this arithmetic at all.

These are genuine wrong clocks, not stale tokens replayed from a retry queue, which would present identically on the wire. Within a single burst the offset stays flat to ±0.5s (stdev 0.3–0.5s) across spans up to 478s, whereas a replayed token's apparent offset grows 1s per elapsed second. One device holds a 170s offset across three separate days.

**Review round (77c3b08)**

Three changes from @aristidesstaffieri's review, all in `77c3b08`:

- **The fail-open predicate is type-based, not string-based.** It keyed on the flattened `auth.Reason(err)` string, but that string is a plain field set at ten `*VerificationError` sites across `internal/auth`, so the fail-open set was "anything whose reason matches" — extensible from another file without touching the middleware. Now `errors.As` on `*ExpiredTokenError` / `*ClockAheadError`. No behavior change today (none of the ten sites carries a clock reason), but extension now requires defining a new error type *and* naming it here. This mechanism had already failed once in this branch: `5b32ac7` permitted `ReasonBadTiming` and served tokens with no `exp` claim at all, caught by review rather than by a test. `TestAuth_PermissiveDoesNotFailOpenOnReasonStringAlone` now pins it — verified to fail (200, handler reached, `invalid_permitted` incremented) when the predicate is reverted.
- **The `iss` metric label uses the signature-verified issuer where one exists.** The permitted fall-through moved that label onto the *serving* path, where `IssuerFromRequestUnverified`'s own contract says not to rely on it. `ExpiredTokenError` now carries `Issuer`, authentic because `ParseWithClaims` verifies the signature before jwt/v5 validates claims (confirmed in v5.2.3: `parser.go:113` precedes `:123`). `ClockAheadError` has no counterpart by construction. The rejection log gains `iss_verified` so the trust level is self-describing rather than inferred from `user_id`'s absence.
- **The permitted path is tested with a request body.** The fall-through retired the invariant "if the verifier touched the body, verification succeeded" — a handler now runs after `parseJWT` returned an error, on a request the verifier already drained. Three of the four endpoints affected in prd are POSTs with bodies, and that was the untested half of the new path. Adds `TestVerifyHTTPRequest_ResetsBodyOnValidationFailure` (`internal/auth`) and `TestAuth_PermittedRequestReachesHandlerWithIntactBody` (both clock directions, `bodyHash` bound to the real body). `authtest.MintToken` now takes an explicit `body` instead of hardcoding `HashBody(nil)` — that default is why every middleware test was nil-body and this gap stayed invisible.

**Metric consumers must be updated**

Anything watching the invalid-token rate must now sum `result="rejected"` with `result="invalid_permitted"`, or permitted skew reads as having vanished. This matters most for permissive→strict flip readiness. Nothing currently alerts on it (re-checked `wallet-eng-runbooks` after `77c3b08` — zero hits for `iss_verified`, `freighter_auth_requests_total`, `invalid_permitted`, `clock_ahead`, `bad_timing`, or the log message), so there is no live alert to fix — but the strict-flip check must be written against the sum.

**The flip gate must keep both timing reasons.** They carry different evidential weight: `invalid_permitted{reason="expired"}` is authentic per-client (verified `iss`), so it is the number for deciding which client team still needs the `Date`-header fix; `reason="clock_ahead"` is an unattributable upper bound, since no signature ran and one increment needs no key. It is tempting to narrow the gate to `expired` for that reason — don't. `clock_ahead` is 14.9% and rising, so an `expired`-only gate would read ready-to-flip while a growing fast-clock population is still broken, trading a fail-safe criterion for a fail-dangerous one.

**Verification**

- `go build ./...` clean; `go test ./...` fully green (19 packages); `go test -race` green on `internal/auth`, `internal/api`, `internal/metrics`; `go vet` and `gofmt` clean. CI was green through `9947b54`; re-running for `77c3b08`.
- Tests written first and confirmed load-bearing: with the new tests and the middleware reverted, `permissive/expired` and `permitted expired` fail; with it restored they pass. `permissive/wrong-key` and every `required/*` row pass in both states, which is what proves the narrowness and that strict is untouched.
- `TestAuth_PermissiveServesEveryObservedProdClockLag` replays every real offset from the prd logs through the middleware and asserts each is served, carries no `userID`, and counts as `invalid_permitted/expired`. This pins the fix to the incident data rather than to a chosen threshold. Now seven rows: the original five plus 19h06m and 475h14m (~19.8 days), which the 7-day window turned up. The 475h row is the most useful one here — no leeway anyone would accept covers it.
- `TestAuth_PermissiveServesArbitrarilyLargeClockOffsetEitherDirection` asserts ±24h/±30d/±365d offsets are served — the property that distinguishes this from widening the leeway, since there is no bound left to re-tune. The fast-clock rows were added when prod had shown only lagging clocks, on the reasoning that five devices cannot support a conclusion that fast clocks do not occur. That reasoning has since been vindicated by data: 78 `clock_ahead` against 442 `expired` over 7 days, and the majority on the two most recent days. They now guard a live population rather than a symmetry argument.
- `TestAuth_PermissiveStillRejectsMalformedTiming` asserts the complement — an over-long lifetime is `bad_timing`, still 401s, and counts as `rejected` rather than permitted skew.
- `internal/api` route-wiring test still asserts 401 in permissive using `not-a-real-token` — that is `malformed`, so it correctly still 401s; comment tightened so it doesn't imply broader coverage than it has.

**Follow-ups / out of scope**

1. **This does nothing for strict**, which stays fail-closed by design. The Blockaid migration (SE-12017) is driving that flip, so this is a rollout-window mitigation with an expiry date.
2. **The real fix is client-side**: derive a clock offset from the `Date` response header and sign with `Date.now() + offset`, re-derived per response so it self-corrects when a clock is fixed. `buildAuthJwt` already takes an injectable `now` and `authedFetch` already has a 401-retry hook (which today rebuilds from the same wrong clock, hence the deterministic double failure).
3. **That fix needs one line of server CORS first**: `Access-Control-Expose-Headers: Date` in `internal/api/middleware/header.go`. `Date` is not CORS-safelisted, and the extension manifest is MV3 with no `host_permissions`, so JS currently reads `null`. Mobile is unaffected (no CORS). Not included here to keep the change reviewable, but it blocks the client work and can ship immediately.
4. **Silent-failure caveat**: this converts a loud failure into a quiet one, which is the right trade only while no gated route consumes identity — true in prd today (`DB_ENABLED=false`, `WALLET_BACKEND_ROUTES_ENABLED=false`), not true once contacts or balances turn on.
5. **Consider a `freighter_auth_clock_skew_seconds` histogram** on timing rejections. The skew magnitude currently exists only in log text, so tracking whether the population is shrinking requires parsing logs rather than reading a dashboard.

</details>


internal/api/middleware/auth.go
```diff
@@ -64,22 +67,52 @@ func Auth(verifier auth.HTTPRequestVerifier, mode auth.Mode, authMetrics *metric
...
+				// Timing reasons ONLY. A bad signature, bad binding, or malformed token
+				// is a real bug or attack rather than a wrong clock, and stays loud.
+				// Strict stays fail-closed for everything: it has no anonymous path to
+				// fall back to, so skewed clients must be fixed client-side before the
+				// permissive→strict flip.
+				//
+				// Grants nothing: no userID is attached, so this request is treated
+				// exactly like one carrying no Authorization header at all.
+				permitted := mode == auth.Permissive &&
+					(reason == auth.ReasonExpired || reason == auth.ReasonBadTiming)
```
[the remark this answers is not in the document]
piyalbasu: Confirmed and fixed in 0c1478b69f57ec4965f1857dcfeaa753c46191d9 — narrowed the predicate to `reason == auth.ReasonExpired` alone.

Reproduced before changing anything: `TestAuth_PermissiveRejectsForgedTokenWithBadTiming` mints a token signed with an attacker key and dated an hour ahead, and against the previous commit it returned **200 with the handler reached** and incremented `invalid_permitted{reason="bad_timing"}`. Exactly as described.

Worth recording that this was not a privilege escalation: the permitted path attaches no `userID`, so the request is equivalent to one carrying no `Authorization` header, which these routes already serve. The real damage was to two contracts I had written — "a bad signature always stays loud" became false, and forged tokens could inflate the counter intended to gate the permissive→strict flip. Your second point is also right and I had missed it independently: `ReasonBadTiming` covers missing `exp`/`iat`, `exp` preceding `iat`, and over-long lifetimes, so it was never a pure clock signal.

`ReasonExpired` is safe by construction, which is why narrowing works: it originates from `jwtgo.ParseWithClaims`, and jwt/v5 returns on signature failure *before* validating claims (`parser.go:113` precedes `:123`), so the classification implies the signature verified.

On your suggested alternative — a signature-verified, skew-specific reason — I did not take it, and want to be explicit about the trade rather than silently declining. It means reordering `parseJWT` to verify the signature before the cheap claim checks, which inverts a deliberate ordering in a security-critical path. The only thing it buys is fast-clock coverage, and prod says that is empty: over the first 24h of real JWT traffic, `bad_timing` was **0** against **335** `expired` (#147). Reworking that path for a zero-occurrence case inside a mitigation PR is the wrong risk to take. The residual gap is now documented as deliberate in the code comment, the truth table, and the design doc, so if fast clocks ever show up the reason is recorded rather than rediscovered.

internal/metrics/metrics.go
```diff
@@ -92,12 +92,23 @@ func SanitizeClient(iss string) string {
 	}
 }
 
+// ResultInvalidPermitted is the RequestsTotal `result` for a token that failed
+// verification on timing alone and was served anonymously rather than rejected
+// (permissive mode only — see middleware.Auth).
```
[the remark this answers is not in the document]
piyalbasu: Fixed in 0c1478b69f57ec4965f1857dcfeaa753c46191d9. The const now counts `reason="expired"` and nothing else, so the contract it documents is true rather than aspirational.

The doc comment states the reason explicitly instead of the vaguer "timing" framing:

> Scope is exactly `reason="expired"`, which is the only verification failure that implies the signature was checked and passed (jwt/v5 returns on signature failure before it validates claims). So every request counted here came from a genuine holder of the subject's private key whose clock was running behind — not from a forged or malformed token. `reason="bad_timing"` is NOT counted here despite also being a clock symptom, because it is classified before signature verification.

This mattered more than the middleware change itself: this counter is what a permissive→strict readiness check reads, so a population that silently included forged and malformed tokens would have made "skew has converged, safe to flip" unfalsifiable.

docs/design/2026-06-22-jwt-auth-middleware.md
```diff
@@ -37,14 +37,34 @@ Shipped Freighter clients today send **no** JWT; newer client versions send a JW
...
-A present-but-invalid token is **always** rejected (401) in both modes — only updated clients send
-tokens, so a bad token is a real bug or attack, and rejecting it gives clean adoption signal during
-rollout.
+| Mode | No `Authorization` header | Header present, valid | Invalid: timing only | Invalid: any other reason |
+| --- | --- | --- | --- | --- |
+| **permissive** (default) | pass (anonymous, no `userID`) | pass (+`userID`) | pass (anonymous, no `userID`) | **401** |
+| **strict** (`auth.Required`) | **401** | pass (+`userID`) | **401** | **401** |
+
+"Timing only" means `expired` or `bad_timing` — the token is well-formed and correctly signed, and
+its only fault is a timestamp from a device clock that disagrees with ours.
```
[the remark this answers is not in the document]
piyalbasu: Fixed in 0c1478b69f57ec4965f1857dcfeaa753c46191d9. The mode table column is now `Invalid: expired` rather than "timing only", and the prose below it explains why that narrowness is load-bearing instead of leaving it as an unexplained choice — including your specific point that `bad_timing` covers missing `exp`/`iat`, `exp < iat`, and excessive lifetime, and that those run ahead of signature verification.

I also added the residual gap explicitly, since the previous table promised more than the code delivered and I would rather it now under-promise visibly:

> **Known gap:** a clock running *fast* by more than the leeway is `bad_timing`, so it still 401s. Accepted deliberately — in the first 24h of real JWT traffic `bad_timing` was 0 against 335 `expired`. Closing it safely would need a signature-verified, skew-specific reason, i.e. reordering `parseJWT` to verify before validating claims, which the data does not justify.

That is the second time this document has promised behavior its rationale did not support — the original "only updated clients send tokens, so a bad token is a real bug or attack" is what produced #147 in the first place. Both are now stated with the counterexample attached.

internal/api/middleware/auth.go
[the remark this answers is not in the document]
piyalbasu: Correction — my reply above is now stale, and the outcome moved toward your suggestion rather than away from it.

I said I'd narrowed to `expired` and declined the skew-specific reason. On review that narrowing was over-corrective: it left a user whose clock runs *fast* locked out for a reason that has nothing to do with them, and my two justifications didn't hold. "A bad signature must stay loud" re-imported reasoning #133 had already discarded (the signal we need goes to metrics/logs, not to the client via a 401). And "`bad_timing` was 0 in prod" was 0 across **five devices** — nowhere near enough to conclude fast clocks don't occur.

What did hold was your *second* point, which turned out to be the load-bearing one: `bad_timing` was never a clock signal. So in dda760f I took half your suggestion — the skew-specific reason — and split `ReasonClockAhead` + `ClockAheadError{AheadBy}` out of `bad_timing` for the two genuine clock branches in `Claims.Validate`. `bad_timing` keeps the malformed cases (missing `exp`/`iat`, `exp` preceding `iat`, over-long lifetime) and still 401s. Permitted set is now `expired` + `clock_ahead`.

I did not take the other half — verifying the signature before permitting. That means `clock_ahead` still precedes signature verification, so a forged token dated into the future is served. Safe rather than tolerated: permitting attaches no `userID`, so the request is byte-for-byte equivalent to one with no `Authorization` header, which these routes already serve — an attacker gains nothing they couldn't get by sending no token at all. `TestAuth_PermissiveForgedTokenOutcomes` pins all three forged-token outcomes and asserts the invariant that makes it safe: a forged token never yields a `userID`.

The residual cost is measurement, not access, and it's now documented at the metric, the middleware, and the design doc: `invalid_permitted{reason="clock_ahead"}` is an upper bound on fast-clock clients, not an exact count. `expired` has no such caveat.

internal/metrics/metrics.go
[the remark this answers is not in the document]
piyalbasu: Correction — my reply above said the scope is "exactly `reason=expired`". As of dda760f it is `expired` **and** `clock_ahead`, the two directions a clock can be wrong.

Your point stands and is what drove the shape of the fix: the counter must not mix wrong clocks with malformed tokens, since it's what a permissive→strict readiness check reads. The answer was to split `bad_timing` rather than to pick a side of it — `clock_ahead` for the two real clock branches, `bad_timing` keeping the malformed cases and still 401ing.

The const doc now spells out the asymmetry between the two permitted reasons rather than implying they're equivalent:

> - `"expired"` — clock lagging. Implies the signature verified, since jwt/v5 returns on signature failure before validating claims. A real user.
> - `"clock_ahead"` — clock fast. Assigned during `Claims.Validate`, i.e. BEFORE signature verification, so it does NOT imply a real user; a forged token dated into the future lands here too. Read it as an upper bound on fast-clock clients, not an exact count.

docs/design/2026-06-22-jwt-auth-middleware.md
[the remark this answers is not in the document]
piyalbasu: Correction — my reply above described a `Invalid: expired` column and a "known gap" for fast clocks. As of dda760f the column is `Invalid: wrong clock` and covers both directions; the gap is closed rather than documented.

The table now carries a per-reason breakdown of exactly the distinction you raised — whether the reason implies a verified signature:

| reason | meaning | signature verified? |
| --- | --- | --- |
| `expired` | clock lagging — `exp` already past | **yes** — raised by `jwtgo.ParseWithClaims`, and jwt/v5 returns on signature failure *before* validating claims |
| `clock_ahead` | clock fast — `iat`/`exp` too far ahead | **no** — raised in `Claims.Validate`, which runs first |

Plus your specific point about what `bad_timing` actually contains (missing `exp`/`iat`, `exp` preceding `iat`, over-long lifetimes), why those stay rejected, and an explicit paragraph on the forged-token consequence of permitting a pre-signature reason.

Worth recording why this doc kept over-promising: the original rationale — "only updated clients send tokens, so a bad token is a real bug or attack" — is what produced #147, and my first fix then under-promised in the opposite direction. Both the rule and its counterexample are now stated inline so the next reader gets the constraint, not just the conclusion.

internal/api/middleware/auth.go
```diff
@@ -64,22 +70,73 @@ func Auth(verifier auth.HTTPRequestVerifier, mode auth.Mode, authMetrics *metric
...
+				// Authorization header, which these routes already serve. An attacker
+				// gains nothing they could not have by sending no token at all. What it
+				// does mean is that invalid_permitted{reason="clock_ahead"} is a noisier
+				// signal than the expired one — read it as an upper bound on fast-clock
+				// clients, not an exact count.
+				//
+				// Strict stays fail-closed for everything: it has no anonymous path to
+				// fall back to, so skewed clients must be fixed client-side before the
+				// permissive→strict flip.
+				permitted := mode == auth.Permissive &&
```
aristidesstaffieri: **Suggestion: make the fail-open predicate type-based rather than string-based.**

No live bug here — I checked all 10 `&VerificationError{...}` constructions in `internal/auth` and none carries `ReasonExpired` or `ReasonClockAhead`, so current behavior is exactly what the PR documents. This is about the property being structural instead of conventional.

Both clock failures have dedicated types, and `Reason()` already discriminates them with `errors.As` before flattening to a string. The predicate then keys on the flattened value, and `Reason()`'s third branch returns `ve.Reason` verbatim — a plain `string` field set at 10 sites across `claims.go`, `parser.go`, `verifier.go`. So the fail-open set isn't "the two clock error types," it's "anything whose reason string matches," which anyone editing `internal/auth` can extend without opening this file.

What makes me want to fix it rather than note it: **the mechanism already produced exactly this failure, in this branch.** `5b32ac7` read

```go
permitted := mode == auth.Permissive &&
    (reason == auth.ReasonExpired || reason == auth.ReasonBadTiming)
```

which served tokens with no `exp` claim at all, and 2h-lifetime tokens, and counted them as `invalid_permitted`. `0c1478b` fixed it — but it was caught by review, not by a test: `git show 5b32ac7:internal/api/middleware/auth_test.go | grep -c bad_timing` → `0`. That commit updated the truth table to assert the new 200s, so the suite encoded the behavior rather than resisting it. `TestAuth_PermissiveStillRejectsMalformedTiming` landed later, in `dda760f`.

And the next one would be equally invisible: three of the four remaining `bad_timing` sub-cases have no test anywhere in the repo (missing `exp`, missing `iat`, `exp` precedes `iat` — only over-long lifetime is covered). Meanwhile the prevailing idiom in `claims.go` is `&VerificationError{Reason: ...}` — 6 of 8 returns — so a new check written in the file's own style is the fail-open-capable form by default. Concretely: `Claims.Validate` doesn't check `nbf`, and a future `nbf` *is* a fast-clock symptom, so whoever adds that check reaches for `ReasonClockAhead` in the surrounding style and has added a pre-signature fail-open path that reads as correct in review.

Worth being clear about the stakes, since the PR is already careful about this: not a privilege escalation either way — no `userID` attaches, and `TestAuth_PermissiveForgedTokenOutcomes` pins that well. The damage is to `invalid_permitted`, which the design doc makes the gate for the strict flip. Contaminating it with malformed tokens is exactly the harm `0c1478b` was written to undo.

Suggested change — four lines, no new concepts, `errors` already imported, and `reason` stays as-is for `RecordAuth` and the log line:

```go
var expired *auth.ExpiredTokenError
var ahead *auth.ClockAheadError
permitted := mode == auth.Permissive &&
    (errors.As(err, &expired) || errors.As(err, &ahead))
```

Extending the fail-open set then requires defining a new error type *and* naming it here — a diff a reviewer actually sees.

internal/api/middleware/skew_regression_test.go
```diff
@@ -0,0 +1,176 @@
...
+			var gotUserID string
+			var hasUser bool
+			next := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				reached = true
+				gotUserID, hasUser = auth.UserIDFromContext(r.Context())
+				w.WriteHeader(http.StatusOK)
+			})
+			handler := Auth(auth.NewVerifier(auth.ClockSkewLeeway), auth.Permissive, m)(next)
+
+			r := httptest.NewRequest(http.MethodGet, authTestPath, nil)
```
aristidesstaffieri, line 83: **Non-blocking: two tests I'd add for the permitted fall-through with a request body.**

The fall-through changes a structural property of this middleware. Before it, `next.ServeHTTP` was reachable from exactly two arms — success (body read and reset) and `ErrNoToken` in permissive (verifier returns at `verifier.go:52`/`:59`, before `readAndResetBody` ever runs). At `7c6eecd`, lines 61/84/95/104 all `return`. So the invariant was: **if the verifier touched the body, verification succeeded.** The permitted `break` retires that — a handler can now run after `parseJWT` returned an error, on a request whose body the verifier already drained.

It's correct today, because `readAndResetBody` (`verifier.go:71`) runs before `parseJWT` (`verifier.go:77`) and unconditionally swaps in a `bytes.Reader` over the full bytes. That ordering is now load-bearing for handler correctness and reads like an implementation detail.

The refactor that would break it is specific and, I'd argue, newly attractive: hoist a cheap timing pre-check above the body read — parse unverified, check `iat`/`exp`, bail before draining. This PR establishes that skewed devices fail ~100% of the time, so "stop draining bodies for the population we know fails timing" is a natural thought. Pre-PR it was safe; those requests 401'd and the body was never needed. Post-PR the handler runs and gets EOF. `TestVerifyHTTPRequest_Valid` stays green through it, because on success timing passes and the body is read as before.

Worth sizing because of which routes ride on it: of the four endpoints your description names as affected in prd, **three are POSTs with bodies** — `token-prices`, `collectibles`, `ledger-key/accounts`. That's the majority of the payload of this change, and it's the untested half of the new path. A regression there surfaces as `io.EOF` inside the handler — a 400, a 500, or an empty-result 200 depending on the handler — and *only* for skewed-clock users, i.e. the population already labeled as broken. Which is exactly the reading that would stop anyone from looking.

**1. `internal/auth` — the one that actually pins the ordering.** Works with the existing helpers as-is, since `skewedClaims` sets `BodyHash: HashBody(body)`:

```go
func TestVerifyHTTPRequest_ResetsBodyOnValidationFailure(t *testing.T) {
	_, priv, sub := newKeypair(t)
	body := []byte(`{"x":1}`)
	// Lagging clock well past the leeway: fails as expired, and methodAndPath /
	// bodyHash both still match so nothing else can be the reason.
	token := mint(t, priv, skewedClaims(sub, "POST /api/v1/thing", body, -(ClockSkewLeeway + time.Minute)))

	r := newRequest(t, http.MethodPost, "/api/v1/thing", body, token)
	_, err := NewVerifier(ClockSkewLeeway).VerifyHTTPRequest(r)
	require.Error(t, err)
	require.Equal(t, ReasonExpired, Reason(err))

	got, err := readAll(r)
	require.NoError(t, err)
	assert.Equal(t, body, got)
}
```

Direct complement to `TestVerifyHTTPRequest_Valid:295`, and it's the test the pre-check hoist would trip.

**2. Middleware level — the end-to-end statement.** Also needs no helper changes, for a slightly subtle reason: `Claims.Validate` returns `clock_ahead` at `claims.go:84`, before the `bodyHash` check at `:113`, so `mintToken`'s `HashBody(nil)` doesn't matter on the fast-clock path.

```go
func TestAuth_PermittedRequestReachesHandlerWithIntactBody(t *testing.T) {
	pub, priv, err := ed25519.GenerateKey(nil)
	require.NoError(t, err)
	sub := hex.EncodeToString(pub)

	body := []byte(`{"addresses":["GABC"]}`)
	token := mintToken(t, priv, sub, "POST "+authTestPath, auth.MaxTokenLifetime, time.Now().Add(time.Hour))

	var got []byte
	next := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		got, _ = io.ReadAll(r.Body)
		w.WriteHeader(http.StatusOK)
	})
	handler := Auth(auth.NewVerifier(auth.ClockSkewLeeway), auth.Permissive, nil)(next)

	r := httptest.NewRequest(http.MethodPost, authTestPath, bytes.NewReader(body))
	r.Header.Set("Authorization", "Bearer "+token)
	rr := httptest.NewRecorder()
	handler.ServeHTTP(rr, r)

	require.Equal(t, http.StatusOK, rr.Code)
	assert.Equal(t, body, got, "a permitted request must reach its handler with an intact body")
}
```

**One observation if you want the `expired` direction covered at the middleware level too:** that path *does* reach the `bodyHash` check, so it needs a token bound to the real body — and `authtest.MintToken` hardcodes `HashBody(nil)` (`authtest.go:24`). A `body []byte` parameter (or a `MintTokenWithBody` sibling) would unblock it, and is probably worth having regardless: that hardcoded nil is why every middleware test is nil-body, which is how this gap stayed invisible across `5b32ac7`, `1f52ffa`, and `dda760f`.

internal/api/middleware/auth.go
```diff
@@ -64,22 +70,73 @@ func Auth(verifier auth.HTTPRequestVerifier, mode auth.Mode, authMetrics *metric
...
+					(reason == auth.ReasonExpired || reason == auth.ReasonClockAhead)
+
+				result := "rejected"
+				if permitted {
+					result = metrics.ResultInvalidPermitted
+				}
+				metrics.RecordAuth(authMetrics, result, reason, metrics.SanitizeClient(iss))
```
aristidesstaffieri, line 165: **The `iss` recorded on this line is unverified, and the flip criterion added in this PR reads it per-client.**

The middleware has two sources for `iss`. The success arm labels from `identity.Issuer`, which is signature-verified (`auth.go:57-60`). This arm labels from `IssuerFromRequestUnverified`. That split used to coincide with serve-vs-reject, which is the assumption baked into the helper's own doc comment: *"use it ONLY for a bounded, best-effort observability label (metric client bucket, log field) on the rejection path — never for an authentication or authorization decision."* The permitted fall-through moves it onto the serving path.

For `expired`, an authentic issuer exists and is discarded. `parseJWT` reaches `ExpiredTokenError` only after `ParseWithClaims` verified the signature, so `claims.Issuer` is authentic at that point. `verifier.go:76` then returns `Identity{}, err`, and this arm re-derives a weaker version by parsing the token a third time. For `clock_ahead` there's no alternative — no signature check ran, and `decodePublicKey` (`parser.go:45`) never ran either, so `sub` isn't even a validated key. That half is irreducibly caller-chosen.

What it affects, in order:

1. **Per-client attribution.** Design doc line 86 gates the flip on "the timing reasons reaching ~0 across both `iss` values." In aggregate that criterion fails safe — an inflated count can only read *not ready*, never ready-when-it-isn't. The per-client split has no such property: it is wrong in whichever direction the sender chose, and it's the number you'd use to decide which client team still needs to ship the `Date`-header offset fix. `SanitizeClient` bounds the label cardinality but not the attribution.

2. **The count is unattributable, at one request per increment.** No key, no signature, no valid `sub` required — `POST /api/v1/token-prices` with a future-dated token claiming any `iss`. The practical cost isn't that someone does this deliberately; it's that the metric alone cannot distinguish five real fast-clock devices from one script, so investigating a nonzero count has to leave the dashboard for logs and source IPs. With the flip on a deadline (SE-12017 per follow-up 1), that's the number the decision rests on.

3. **Log ambiguity on 200s.** `iss` now means "verified" for authenticated requests and "spoofable" for permitted ones, with `user_id`'s absence as the only tell. An aggregation by `iss` filtered to 200s mixes the two.

`expired` was 100% of the real population — the regression test records 335 expired against 0 fast-clock — so recovering the verified issuer for that half yields an authentic per-`iss` signal covering every observed device, and leaves `clock_ahead` as the explicitly noisy bucket. Line 86 could then gate on `invalid_permitted{reason="expired"}` per-`iss` instead of on a blended number that inherits the weaker guarantee.

Suggested change — mirror how `ExpiredBy` already carries diagnostic detail out of the failure:

```go
type ExpiredTokenError struct {
	ExpiredBy time.Duration
	Issuer    string // signature-verified: ParseWithClaims checks the signature before claims
	Err       error
}
```

`claims` is in scope where it's constructed in `parseJWT`, so it's a one-word addition — and deliberately not a signature change to `VerifyHTTPRequest`, since returning a partial `Identity` on error would invite reading `UserID` off it. Then here:

```go
iss := auth.IssuerFromRequestUnverified(r)
var expired *auth.ExpiredTokenError
if errors.As(err, &expired) && expired.Issuer != "" {
	iss = expired.Issuer
}
```

If you take the type-based predicate suggested on line 117, that `errors.As` is already on this path, so this comes nearly free on top.

For the log field, the cheapest fix is a distinct name on the permitted path — `iss_unverified` rather than `iss`, or a `verified` boolean beside it — so the trust level is self-describing rather than inferred from `user_id`'s absence.

internal/api/middleware/auth.go
piyalbasu: Taken in 3239f30adf2ccc533cece8f35f0b687e9a07268e — the predicate now discriminates on the error types.

**Confirmed your no-live-bug finding independently:** none of the ten `&VerificationError{...}` sites carries a clock reason, so behavior today is unchanged. What sold the change is that the mechanism has already failed once in this branch, and the `nbf` example is the version I couldn't argue with — `Claims.Validate` genuinely doesn't check it, a future `nbf` genuinely is a fast-clock symptom, and the surrounding style genuinely hands you the fail-open-capable form.

**One thing I added beyond the suggestion:** the property is now enforced by a test rather than by the type system alone.

<details>

**The change**, as suggested:

```go
var expiredErr *auth.ExpiredTokenError
var clockAheadErr *auth.ClockAheadError
isExpired := errors.As(err, &expiredErr)
isClockAhead := errors.As(err, &clockAheadErr)
...
permitted := mode == auth.Permissive && (isExpired || isClockAhead)
```

`reason` is still computed and still feeds `RecordAuth` and the log line, so the metric/log surface is byte-identical.

**New test:** `TestAuth_PermissiveDoesNotFailOpenOnReasonStringAlone`. It hands the middleware a `stubVerifier` returning `&auth.VerificationError{Reason: auth.ReasonExpired}` — a clock reason *string* without the clock error *type* — and asserts 401, handler not reached, `result="rejected"`, and zero `invalid_permitted`. Table-driven over both clock reasons.

`Auth` already takes the `auth.HTTPRequestVerifier` interface, so this needed no production seam.

**Verified the test actually resists the regression** rather than merely passing: reverting the predicate to the string form makes it fail with `expected: 401, actual: 200`, "handler must not be reached", and `invalid_permitted` incremented — i.e. it reproduces exactly the `5b32ac7` failure mode you identified.

Your point that three of the four remaining `bad_timing` sub-cases are untested is separately true, and I've left it alone deliberately — `TestAuth_PermissiveStillRejectsMalformedTiming` covers the bucket's *outcome* at the middleware boundary, which is what this predicate can affect. Per-branch coverage of `Claims.Validate` is a `internal/auth` concern and a cleaner standalone change than riding along here.

</details>

internal/api/middleware/skew_regression_test.go
piyalbasu: Both tests added in 3239f30adf2ccc533cece8f35f0b687e9a07268e, plus the `authtest` change you suggested. The structural point is right and was the untested half of the change.

**One correction:** the specific refactor you named as the breaking one doesn't actually break it. Hoisting a timing pre-check *above* the body read means the body is never drained, so the handler still gets it intact — I simulated it and the proposed test passes. The regression that does break it is draining and then resetting only on the success path. Same invariant, different trigger; the test is written against the one that bites.

<details>

**On the mechanism.** I built your hypothesised refactor — parse unverified, check `iat`/`exp`, bail before `readAndResetBody` — and `TestVerifyHTTPRequest_ResetsBodyOnValidationFailure` passed. Early-returning before the read leaves `r.Body` untouched, so there's no `io.EOF` for the handler to hit.

The regression that *does* trip it: read the body, then restore it only after `parseJWT` succeeds. Simulated that and the test fails with `expected: []byte{0x7b,0x22,0x78,...}, actual: []byte{}` while `TestVerifyHTTPRequest_Valid` stays green — the asymmetry you were after. That's now what the test comment describes.

**Test 1** — `TestVerifyHTTPRequest_ResetsBodyOnValidationFailure` in `internal/auth`, essentially as you wrote it.

**Test 2** — `TestAuth_PermittedRequestReachesHandlerWithIntactBody`, extended to a table over **both** clock directions. Your version used the fast-clock path, which passes even with a mismatched `bodyHash` because `Claims.Validate` returns `clock_ahead` before reaching the `bodyHash` check (`claims.go:81` vs `:113`). That means the fast-clock row alone never exercises the bodyHash-verified path — so the `expired` row is the one carrying the weight here, and it needs a token bound to the real body.

**`authtest.MintToken` now takes an explicit `body []byte`** rather than a `MintTokenWithBody` sibling. A sibling leaves the nil default in place as the path of least resistance, which is the trap you identified; a required parameter removes it. Churn was two direct callers — the 16 sites going through the local `mintToken` wrapper are untouched, and there's a `mintTokenWithBody` wrapper beside it.

**Verified:** `go test -race ./internal/auth/... ./internal/api/...` green; full suite green.

</details>

internal/api/middleware/auth.go
piyalbasu: `ExpiredTokenError.Issuer` added in 3239f30adf2ccc533cece8f35f0b687e9a07268e, exactly as suggested, plus `iss_verified` on the log line. The unverified-label-on-the-serving-path problem is real and the fix is as cheap as you said.

**But the premise the recommendation rests on has expired, and it inverts your point 1.** You wrote that `expired` was 100% of the real population, so gating per-`iss` on `reason="expired"` would cover every observed device. That was true of the 24h window in the regression test; it is not true now. Over the 7 days ending 2026-08-03, prd logged **442 `expired` (84.4%) vs 78 `clock_ahead` (14.9%)** — and `clock_ahead` was the **majority** on the two most recent days (08-02: 38 vs 44; 08-03: 20 vs 8).

So an `expired`-only gate is now **fail-dangerous**: it would read "ready to flip" while the growing fast-clock population is still broken. I've kept both reasons in the gate and documented the attribution asymmetry instead.

<details>

**Taken as suggested.** `ExpiredTokenError` gains `Issuer string`, populated at `parser.go` where `claims` is already in scope. Your reasoning checks out: `ErrTokenExpired` is only reachable after `ParseWithClaims` verified the signature — confirmed in jwt/v5 v5.2.3 itself, `parser.go:113` returns `ErrTokenSignatureInvalid` before `p.validator.Validate(claims)` at `:123`. `ClockAheadError` deliberately has no counterpart, documented on the type.

Middleware, as you wrote it, folded into the `errors.As` from the line-117 comment so it costs nothing extra:

```go
iss := auth.IssuerFromRequestUnverified(r)
issVerified := false
if isExpired && expiredErr.Issuer != "" {
    iss = expiredErr.Issuer
    issVerified = true
}
```

**Log field:** went with the `verified` boolean rather than renaming to `iss_unverified`, because `iss` is what existing Loki queries and the runbook greps key on. `iss_verified` is set both as a context field and on the rejection line, so a filtered aggregation is a predicate rather than a rename.

New test `TestAuth_PermittedExpiredUsesSignatureVerifiedIssuer` signs a token with `iss=freighter-mobile`, lagging past the leeway, and asserts the counter lands on `freighter-mobile` and that `ExpiredTokenError.Issuer` carries it.

**On the flip criterion (your point 1).** Design doc now states the asymmetry explicitly:

- `invalid_permitted{reason="expired"}` — authentic per-client, use it to decide *which client team* still needs the `Date`-header offset fix.
- `invalid_permitted{reason="clock_ahead"}` — upper bound, and unattributable: no signature ran, one increment needs no key, so do **not** split it by client.

...and then says not to narrow the gate to `expired` despite that, with the numbers above as the reason. Your framing was that aggregate fails safe while per-client doesn't; the data adds a third case — *reason-narrowed* also fails dangerous, and in the direction that matters, because the noisier bucket is the one that's growing.

**Also refreshed the stale prod claims** this comment surfaced, in the same commit: the `observedProdClockLags` header, the fast-clock note on `...ArbitrarilyLargeClockOffsetEitherDirection`, and the design doc all previously asserted fast clocks were 0. Added the two lags the 7d window turned up that the table was missing (19h06m, 475h14m ≈ 19.8 days).

**One thing worth recording**, since it's the kind of claim that looks unfalsifiable: these are genuine wrong clocks, not stale tokens replayed from a retry queue — which would present identically on the wire. Within a single burst the offset stays flat to ±0.5s across spans up to 478s (stdev 0.3–0.5s), whereas a replayed token's apparent offset grows 1s per elapsed second. Offsets are also stable per device across days — one device sits at exactly 170s on three separate days.

</details>

aristidesstaffieri:
## `auth.Reason` conflates two different classes of fact

`Reason` is used as a single flat enum, but its values come from two places that differ in whether an attacker controls them:

- **Pre-signature** (`internal/auth/claims.go`, reached at `parser.go:40`) — `bad_timing`, `clock_ahead`, `bad_method_path`, `bad_body_hash`. These are assertions read out of an **unverified** token. Anyone can produce any of them.
- **Post-signature** (`parser.go:49-66`) — `expired`, `bad_signature`. These are facts about a token whose signature verified (or didn't).

`parser.go:40` runs `claims.Validate` before `decodePublicKey` (`:44`) and `ParseWithClaims` (`:49`), so nothing in the first group carries proof of key control. Three consumers treat the two groups as interchangeable.

### 1. The permit predicate spans both groups

```go
permitted := mode == auth.Permissive &&
    (reason == auth.ReasonExpired || reason == auth.ReasonClockAhead)
```

`middleware/auth.go:117-118` — `expired` is signature-backed; `clock_ahead` is not.

The comment at `:100-112` sees this and argues it's safe because permitting attaches no userID, so the request is equivalent to sending no `Authorization` header. That's correct as far as the *response* goes. But it doesn't hold for the metric, and the comment 20 lines earlier makes exactly the opposite argument to exclude `bad_timing`:

> serving them would be indistinguishable in the metrics from serving genuine skew, **in the exact counter that gates the permissive→strict flip**

`clock_ahead` has that property *more* strongly than `bad_timing` does — both are pre-signature, and `clock_ahead` is the one being read as the skew distribution. The `:110-112` mitigation ("read it as an upper bound") is the whole problem restated: an attacker sets the upper bound, and the rollout gate is a comparison against ~0.

It's worse than an unbounded count, because `client` is attacker-chosen too. `:79` takes `iss` from `IssuerFromRequestUnverified`, and `:124` feeds it to `SanitizeClient` on the permitted path — so the attacker picks which client bucket the fabricated skew lands in. A flood of `iat = now+1h` tokens with `iss: "freighter-extension"` and a random signature never reaches `ParseWithClaims` and is indistinguishable in the metric from real extension users with fast clocks.

### 2. The defense-in-depth exp bound now fails open

`claims.go:105` was changed from `VerificationError{ReasonBadTiming}` to `ClockAheadError`. Its own comment says the branch is unreachable today and is retained because "it stops being unreachable the moment either preceding check is reordered, loosened, or removed."

Routing it into the permitted set inverts what it's for. If a future change loosens the `iat` check, a token with `exp = now + 10 years` reaches this branch and is now **served 200 and counted as skew** instead of hard-401ed. The branch exists to catch a lifetime-cap bypass; in the scenario it was written for, it now hides it.

### 3. The post-signature side collapses too

Same layering, opposite direction: `parser.go:57` special-cases only `ErrTokenExpired`, so every other `ParseWithClaims` error becomes `bad_signature` at `:65`. `Claims` embeds `RegisteredClaims`, and `Validate` never checks `nbf` — so a future-dated `nbf` fails inside jwt/v5 as `ErrTokenNotValidYet` and is reported as a forged signature. A client whose clock is fast gets a 401 even in permissive (the lockout this PR exists to fix), filed under the counter operators read as "someone is forging tokens."

*(Whether Freighter clients actually emit `nbf` today is unverified from this repo — if none do, this is latent rather than live.)*

### Suggested fix

Make the signature the first gate, so every reason downstream of it is signature-backed:

```
ParseUnverified → decodePublicKey → ParseWithClaims → claims.Validate
```

Then `clock_ahead` means "a real key holder's clock is fast," the permit predicate is uniformly signature-backed, `invalid_permitted` becomes a metric only key holders can move, and `:105` fails closed again for unsigned tokens. Cost is one Ed25519 verify (~50µs) on tokens that currently short-circuit — worth it, but it does mean malformed tokens get more expensive, so it's worth a look at whether that matters at your rate limits.

If the reorder can't land in this PR, the stopgap is to narrow the permit to `expired` only (the state at `0c1478b`) and revert `claims.go:105` to `ReasonBadTiming`. That keeps the gating metric trustworthy, but it re-opens #147 for fast-clock clients — so it's a hold-the-line measure until the reorder lands, not a resting place.

aristidesstaffieri:
> ## `auth.Reason` conflates two different classes of fact
> `Reason` is used as a single flat enum, but its values come from two places that differ in whether an attacker controls them:
> 
> * **Pre-signature** (`internal/auth/claims.go`, reached at `parser.go:40`) — `bad_timing`, `clock_ahead`, `bad_method_path`, `bad_body_hash`. These are assertions read out of an **unverified** token. Anyone can produce any of them.
> * **Post-signature** (`parser.go:49-66`) — `expired`, `bad_signature`. These are facts about a token whose signature verified (or didn't).
> 
> `parser.go:40` runs `claims.Validate` before `decodePublicKey` (`:44`) and `ParseWithClaims` (`:49`), so nothing in the first group carries proof of key control. Three consumers treat the two groups as interchangeable.
> 
> ### 1. The permit predicate spans both groups
> ```go
> permitted := mode == auth.Permissive &&
>     (reason == auth.ReasonExpired || reason == auth.ReasonClockAhead)
> ```
> 
> `middleware/auth.go:117-118` — `expired` is signature-backed; `clock_ahead` is not.
> 
> The comment at `:100-112` sees this and argues it's safe because permitting attaches no userID, so the request is equivalent to sending no `Authorization` header. That's correct as far as the _response_ goes. But it doesn't hold for the metric, and the comment 20 lines earlier makes exactly the opposite argument to exclude `bad_timing`:
> 
> > serving them would be indistinguishable in the metrics from serving genuine skew, **in the exact counter that gates the permissive→strict flip**
> 
> `clock_ahead` has that property _more_ strongly than `bad_timing` does — both are pre-signature, and `clock_ahead` is the one being read as the skew distribution. The `:110-112` mitigation ("read it as an upper bound") is the whole problem restated: an attacker sets the upper bound, and the rollout gate is a comparison against ~0.
> 
> It's worse than an unbounded count, because `client` is attacker-chosen too. `:79` takes `iss` from `IssuerFromRequestUnverified`, and `:124` feeds it to `SanitizeClient` on the permitted path — so the attacker picks which client bucket the fabricated skew lands in. A flood of `iat = now+1h` tokens with `iss: "freighter-extension"` and a random signature never reaches `ParseWithClaims` and is indistinguishable in the metric from real extension users with fast clocks.
> 
> ### 2. The defense-in-depth exp bound now fails open
> `claims.go:105` was changed from `VerificationError{ReasonBadTiming}` to `ClockAheadError`. Its own comment says the branch is unreachable today and is retained because "it stops being unreachable the moment either preceding check is reordered, loosened, or removed."
> 
> Routing it into the permitted set inverts what it's for. If a future change loosens the `iat` check, a token with `exp = now + 10 years` reaches this branch and is now **served 200 and counted as skew** instead of hard-401ed. The branch exists to catch a lifetime-cap bypass; in the scenario it was written for, it now hides it.
> 
> ### 3. The post-signature side collapses too
> Same layering, opposite direction: `parser.go:57` special-cases only `ErrTokenExpired`, so every other `ParseWithClaims` error becomes `bad_signature` at `:65`. `Claims` embeds `RegisteredClaims`, and `Validate` never checks `nbf` — so a future-dated `nbf` fails inside jwt/v5 as `ErrTokenNotValidYet` and is reported as a forged signature. A client whose clock is fast gets a 401 even in permissive (the lockout this PR exists to fix), filed under the counter operators read as "someone is forging tokens."
> 
> _(Whether Freighter clients actually emit `nbf` today is unverified from this repo — if none do, this is latent rather than live.)_
> 
> ### Suggested fix
> Make the signature the first gate, so every reason downstream of it is signature-backed:
> 
> ```
> ParseUnverified → decodePublicKey → ParseWithClaims → claims.Validate
> ```
> 
> Then `clock_ahead` means "a real key holder's clock is fast," the permit predicate is uniformly signature-backed, `invalid_permitted` becomes a metric only key holders can move, and `:105` fails closed again for unsigned tokens. Cost is one Ed25519 verify (~50µs) on tokens that currently short-circuit — worth it, but it does mean malformed tokens get more expensive, so it's worth a look at whether that matters at your rate limits.
> 
> If the reorder can't land in this PR, the stopgap is to narrow the permit to `expired` only (the state at `0c1478b`) and revert `claims.go:105` to `ReasonBadTiming`. That keeps the gating metric trustworthy, but it re-opens #147 for fast-clock clients — so it's a hold-the-line measure until the reorder lands, not a resting place.

this is fixed in 7f9bd7b431283a7290bd820fb9ed95f7c40b55fe

docs/design/2026-06-22-jwt-auth-middleware.md
```diff
@@ -37,14 +37,79 @@ Shipped Freighter clients today send **no** JWT; newer client versions send a JW
...
+| --- | --- | --- | --- | --- |
+| **permissive** (default) | pass (anonymous, no `userID`) | pass (+`userID`) | pass (anonymous, no `userID`) | **401** |
+| **strict** (`auth.Required`) | **401** | pass (+`userID`) | **401** | **401** |
+
+"Wrong clock" is exactly two reasons, the two directions a clock can be wrong:
+
+| reason | meaning | signature verified? |
+| --- | --- | --- |
+| `expired` | clock lagging — `exp` already past | **yes** — raised by `jwtgo.ParseWithClaims`, and jwt/v5 returns on signature failure *before* validating claims |
+| `clock_ahead` | clock fast — `iat`/`exp` too far ahead | **no** — raised in `Claims.Validate`, which runs first |
```
JiahuiWho, line 50: This describes the earlier revision. After the reorder in `parser.go`, the signature is verified before `Claims.Validate` runs, so `clock_ahead` now implies a verified signature

docs/design/2026-06-22-jwt-auth-middleware.md
```diff
@@ -37,14 +37,79 @@ Shipped Freighter clients today send **no** JWT; newer client versions send a JW
...
+merely having shipped, since client rollout lags by days.
+
+The two reasons carry different evidential weight, and the criterion must not be narrowed to the
+stronger one:
+
+- `invalid_permitted{reason="expired"}` is **authentic per-client**. `ExpiredTokenError` carries the
+  signature-verified `iss` out of the failure, and the middleware labels from it rather than
+  re-parsing the token unverified — so this is the number to use when deciding *which client team*
+  still needs to ship the offset fix.
+- `invalid_permitted{reason="clock_ahead"}` is an **upper bound, and unattributable**. No signature
```
JiahuiWho, line 103: Same nit as the table above

internal/metrics/metrics.go
```diff
@@ -92,13 +92,46 @@ func SanitizeClient(iss string) string {
...
+//   - "expired"     — clock lagging. Implies the signature verified, since jwt/v5
+//     returns on signature failure before validating claims. A real user, and the
+//     `client` label is the signature-verified iss (carried out of the failure by
+//     ExpiredTokenError), so per-client attribution on this reason is authentic.
+//   - "clock_ahead" — clock fast. Assigned during Claims.Validate, i.e. BEFORE
+//     signature verification, so it does NOT imply a real user; a forged token
+//     dated into the future lands here too. Read it as an upper bound on
+//     fast-clock clients, not an exact count — and note the `client` label here is
+//     caller-chosen, so do NOT split this reason by client to decide which app
+//     still needs fixing. One request per increment needs no key.
```
JiahuiWho: This contradicts the middleware comment in `auth.go` and the final `parser.go` ordering?

JiahuiWho (approved):
LGTM

PR #149, opened 2026-08-01
feat(api): expose Date CORS response header for clock skew offset (#147)

### LINKED ISSUE #147: Wall clock skew causes degraded experience

namdamdoi68-oss:
Exposes the Date response header in CORS middleware to allow clients to calculate server clock offset and correct wall clock skew. Fixes #147.


namdamdoi68-oss:
Hi @aditya1702 @aristidesstaffieri — PR is ready for review. All CI tests pass cleanly.

